PR #4, opened 2019-03-26
Implementation of S3-stored html templates

benbart:


lambda/common.py
```diff
@@ -31,6 +31,9 @@
 active_sessions = {}
 session_store = os.getenv('SESSION_STORE', 'DB')
 sessttl = int(os.getenv('SESSION_TTL', '168')) * 60 * 60
+html_template_status = ''
+html_template_local_cachedir = '/tmp/templates/' # Should this be settable via ENV?
```
bbuechler: Nah, I think thats fine.

lambda/common.py
```diff
@@ -685,9 +688,39 @@ def refresh_user_profile(user_id):
...
+    except FileExistsError:
+        # good.
+        log.debug('somehow, {} exists already'.format(html_template_local_cachedir))
+    bucket = os.getenv('CONFIG_BUCKET')
+    templatedir = os.getenv('HTML_TEMPLATE_DIR')
+    if not templatedir[-1] == '/': #we need a trailing slash
+        templatedir = '{}/'.format(templatedir)
+
+    client = boto3.client('s3')
+    result = client.list_objects(Bucket=bucket, Prefix=templatedir, Delimiter='/')
```
bbuechler: It might be useful to wrap this whole section in a big `try:` block and return a helpful error message if you can't list+download. Otherwise I see some one screwing up the config and then struggling to troubleshoot it a fat-fingered bucket prefix.

PR #8, opened 2019-04-02
Implemented locally created download role if not provided in params

benbart:


cloudformation/thin-egress-app.yaml
```diff
@@ -207,6 +208,35 @@ Resources:
       Roles:
         - !Sub "${AWS::StackName}-EgressLambdaIamRole"
 
+  DownloadRoleLocal:
+    Type: AWS::IAM::Role
```
bbuechler: We should only create the IAM Role (and policy!) if we need it 

`Condition: CreateDownloadRole`

you can see an example here: 
https://github.com/asfadmin/rain-egress-app/blob/devel/cloudformation/ecs_stack.yaml#L269

benbart: oops, forgot it

PR #16, opened 2019-05-14
WIP: mods to support asf-urs

benbart:


lambda/app.py
```diff
@@ -5,9 +5,9 @@
 from urllib.parse import urlparse, quote_plus
 
 
-from common import get_log, do_auth, get_yaml_file, process_varargs, \
+from common.common import get_log, do_auth, get_yaml_file, process_varargs, \
```
bbuechler: I'm not in love with this sort of syntax. datetime.datetime screws me up all the time.

benbart: I considered our plans for splitting out and up the common part and this is temporary.

PR #19, opened 2019-05-20
WIP: Getting TEA to HEAD binary and text files

benbart:


lambda/app.py
```diff
@@ -298,25 +292,24 @@ def try_download_head(bucket, filename):
...
-    #return 'Finish this thing'
-
+    
     response_headers = {'Content-Type': download['ContentType']}
+
     for header in download['ResponseMetadata']['HTTPHeaders']:
         name = header_map[header] if header in header_map else header
         value = download['ResponseMetadata']['HTTPHeaders'][header] if header != 'server' else 'egress'
         log.debug("setting header {0} to {1}.".format(name, value))
-        response_headers['name'] = value
```
bbuechler: Woah... This seems like it would have been causing all sorts of problems.

benbart: Yep. Just occurred to me that we should check fat egress for this.

PR #25, opened 2019-05-21
[PR-1937] Make TEA optionally NGAP2 compliant

bbuechler:


cloudformation/thin-egress-app.yaml
```diff
@@ -43,10 +42,33 @@ Metadata:
...
+        Parameters:
+          - PrivateVPC
+          - VPCSecurityGroupIDs
+          - VPCSubnetIDs
+          - PermissionsBoundaryName
 
 
 Parameters:
 
+  VPCSecurityGroupIDs:
```
benbart: do we want to emphasize that this is NGAP only?

PR #40, opened 2019-05-30
Add ability to build thin-egress-app-code.zip using Docker

yjpa7145:
This PR adds a `./bin/build.sh` script which will build `pkg/thin-egress-app-code.zip` inside of a Docker container.  This ensures that users building the package will not have to worry about installing build-time dependencies.

### DIFF
--- a/bin/build.sh
+++ b/bin/build.sh
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+docker run \
+  --rm \
+  --mount type=bind,source="$(pwd)",target=/source \
+  python:3-alpine \
+  /source/bin/docker-build.sh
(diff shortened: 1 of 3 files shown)


bin/build.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+docker run \
+  --rm \
+  --mount type=bind,source="$(pwd)",target=/source \
+  python:3-alpine \
```
benbart: I've run into some edge cases (mainly involving crypto) where building in non-`amazonlinux` images causes "module not found" type errors.

bbuechler:
Abandoned.

PR #44, opened 2019-05-31
More Linting for CodeFactor integration

bbuechler:


lambda/common/common.py
```diff
@@ -200,7 +200,7 @@ def do_auth(code, redirect_url):
         log.debug('url: {}'.format(url))
         log.debug('post_data: {}'.format(post_data))
 
-        response = urllib.request.urlopen(post_request)
+        response = urllib.request.urlopen(post_request)                               #nosec URL is *always* URS.
```
benbart: What was its objection here?

PR #50, opened 2019-06-19
WIP Add CF-based Terraform deployment

yjpa7145:


build/Jenkinsfile
```diff
@@ -61,7 +61,6 @@ pipeline {
 
         // Swap in BUILD#
         sh "echo '>> Build Stamping....'"
-        sh 'sed -i -e "s/<BUILD_ID>/${NEWTAG}/" ./cloudformation/thin-egress-app.yaml'
```
yjpa7145: Converted the build id into a CloudFormation parameter, which makes it ~easier~ possible to set from Terraform.

cloudformation/thin-egress-app.yaml
```diff
@@ -42,42 +43,51 @@ Metadata:
...
+
   PrivateVPC:
     Type: String
     Default: ""
     Description: "Optional internal VPC."
 
   LambdaCodeS3Key:
     Type: String
-    Default: "thin-egress-app/tea-code-<BUILD_ID>.zip"
+    Default: ""
```
yjpa7145: Because `BuildId` is now coming in as a CloudFormation parameter, we can't use the build id as part of the default key.  Line 371 of this file fixes that.

build/Jenkinsfile
bbuechler: So this is perma-stamping the CF template, linking it to a specific build and specific zip file.  Changing it to to be a param as you have below breaks that direct connection. 

I think you're trying to avoid having "<BUILD_ID>" in the cloudformation template, but in reality, its best practice (if only for this project) to pull the CF template not from GITHUB, but from the S3 bucket.

bbuechler: I think there may be confusion over what "build_id" is used for. Its intended *not* to be set. Its a stamp in the code so that if you're looking at logs (https://github.com/asfadmin/thin-egress-app/blob/devel/lambda/common/common.py#L76) you know what version of the code you should be troubleshooting.

bbuechler: So, if your logs say you're looking at "build.12", you can actually go see that code in time: https://github.com/asfadmin/thin-egress-app/tree/tea-build.12

cloudformation/thin-egress-app.yaml
```diff
@@ -42,42 +43,51 @@ Metadata:
...
 Parameters:
+  BuildId:
+    Type: String
+    Default: "build.14"
+    Description: The build of the Thin Egress App to be deployed
+
+  DomainName:
+    Type: String
+    Default: ""
+    Description: Custom domain name used by redirect_url
```
bbuechler: Is this for mapping a CNAME DNS entry ontop of the API Gateway url?

bbuechler: Oh, I see. It looks like its some dev/testing apparatus Ben built in. I *think* there might be better ways to manage this natively in API Gateway? I'm not 100% sure.

yjpa7145: Where this comes in handy for us is setting up the ssh forwards that we need to use when testing with the bastion host.  It lets us add a port to the URL.  It may be worth considering adding an optional "port" parameter that would typically default to `443`.

bbuechler: Ideally, in the future, since this an user-facing download app, you'll just need NGAP to wire up their "enterprise" solution in front. With fat egress, the hostname (and protocol, and port) gets properly sorted out by NGINX & python Flask/Requests with the proper PROXY headers... Unfortunately that may not play out in API GATEWAY. You won't know the Cloudfront host name when you create a stack.... Something for use to consider in the future.

PR #56, opened 2019-07-17
New rain-api-core and 'API' as API Gateway StageName

benbart:


build/Jenkinsfile
```diff
@@ -64,6 +64,8 @@ pipeline {
 
         // Swap in BUILD#
         sh "echo '>> Build Stamping....'"
+        sh 'sed -i -e "s/asf.public.code/${CODE_DIR}/" ./cloudformation/thin-egress-app.yaml'
+        sh 'sed -i -e "s/tea-code-<BUILD_ID>.zip/${CODE_ARCHIVE_FILENAME}/" ./cloudformation/thin-egress-app.yaml'
```
bbuechler: This certainly works... On a scale of 1-10, how opposed are you to making this a _little_ less prone to mistake by changing `tea-code-<BUILD_ID>` to `<CODE_ARCHIVE_FILENAME>` in both this file and the template?

benbart: My opposition would be under "5". My thinking was to make it easy to change the archive to different versions by giving less to change. Nice thought, but in practice, I haven't been doing much of that.

bbuechler: Alternatively, **also** replace `tea` with `<APPTAG>` and `sed` that, BUT then its basically doing the exact same thing as this line: `CODE_ARCHIVE_FILENAME="${APPTAG}-code-${NEWTAG}.zip". Ugly duplication. Either way,  since `tea-code` is default **wrong**, I'd like to get it out of there.

benbart: All that's already on its way out

build/Jenkinsfile
```diff
@@ -64,6 +65,8 @@ pipeline {
 
         // Swap in BUILD#
         sh "echo '>> Build Stamping....'"
+        sh 'sed -i -e "s/asf.public.code/${CODE_DIR}/" ./cloudformation/thin-egress-app.yaml'
```
bbuechler: Is this extraneous if you add the `CODE_ARCHIVE_PATH_FILENAME="${CODE_DIR}/${CODE_ARCHIVE_FILENAME}"` above?

bbuechler: Nevermind.

PR #74, opened 2019-09-12
In-region download role and JWT

benbart:
This should not be merged before https://github.com/asfadmin/rain-api-core/pull/8

### DIFF
--- a/cloudformation/thin-egress-app.yaml
+++ b/cloudformation/thin-egress-app.yaml
@@ -531,7 +531,7 @@ Resources:
         - !Ref EgressLambdaDependencyLayer
       #MemorySize: 128
 
-  AsfUrsDomainName:
+  ExtDomainName:
     Type: AWS::ApiGateway::DomainName
     Condition: DomainNameIsSet
     DependsOn:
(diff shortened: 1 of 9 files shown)


cloudformation/thin-egress-app.yaml
```diff
@@ -9,18 +9,27 @@ Metadata:
...
           - Maturity
           - ConfigBucket
           - HtmlTemplateDir
           - StageName
           - EnableApiGatewayLogToCloudWatch
+      -
+        Label:
+          default: Domain Settings
+        Parameters:
+          - DomainName
```
bbuechler: Is this in App Setting AND Domain settings?

cloudformation/thin-egress-app.yaml
```diff
@@ -60,6 +71,16 @@ Parameters:
     Default: ""
     Description: Custom domain name used by redirect_url
 
+  CookieDomain:
+    Type: String
+    Default: ".asf.alaska.edu"
```
bbuechler: I think we should generalize this back to something like `*.project.cloud.earthdata.nasa.gov`.

cloudformation/thin-egress-app.yaml
```diff
@@ -60,6 +71,16 @@ Parameters:
...
+    Description: "The cookies get set to this domain, and any subdomains."
+
+  DomainCertArn:
+    Type: String
+    Default: 'arn:aws:acm:us-east-1:765666652335:certificate/9f6cc382-c06c-4304-81a7-35a3c330f173'
```
bbuechler: Is this optional? I seem to recall that in NGAP, cert management is expected to be a function of NGAP network services and would be part of the CloudFront endpoint rather than anything running INSIDE the internal VPC.

benbart: I'm afraid I know nothing about how domain certs will work with NGAP and cloudfront

cloudformation/thin-egress-app.yaml
benbart: As a default value, you mean?

Issue #85, opened 2019-12-10
When running behind cloudfront, STAGE is not suppressed on redirect


bbuechler:
https://github.com/asfadmin/thin-egress-app/blob/888eb822f616fba75cda7694158784d910742939/lambda/app.py#L66

After URS Auth, the return URL is from https://d24c78m3es2lu9.cloudfront.net/GRD/S5/SIRC1_16_GRD_006_010_010_19940409T185535_19940409T185550.zip is https://d24c78m3es2lu9.cloudfront.net/API/ GRD/S5/SIRC1_16_GRD_006_010_010_19940409T185535_19940409T185550.zip

Issue #84, opened 2019-12-10
In NGAP, behind CloudFront, STAGE is not suppressed, creating invalid url


bbuechler:
https://github.com/asfadmin/thin-egress-app/blob/724ced707c8dd356725a2ad7152c1d137447a941/lambda/templates/root.html#L16

https://d24c78m3es2lu9.cloudfront.net/API/logout instead of https://d24c78m3es2lu9.cloudfront.net/logout

bbuechler:
This was done

PR #86, opened 2019-12-10
NGAP/Cloudfront Integration changed to mask STAGE

bbuechler:
Resolves https://github.com/asfadmin/thin-egress-app/issues/85 and https://github.com/asfadmin/thin-egress-app/issues/84

### DIFF
--- a/lambda/app.py
+++ b/lambda/app.py
@@ -84,7 +84,7 @@ def make_redriect(to_url, headers=None, status_code=301):
 
 
 def make_html_response(t_vars:dict, hdrs:dict, status_code:int=200, template_file:str='root.html'):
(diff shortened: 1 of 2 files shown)


lambda/app.py
```diff
@@ -63,7 +63,9 @@ def restore_bucket_vars():
 def do_auth_and_return(ctxt):
 
     log.debug('context: {}'.format(ctxt))
-    here = ctxt['path']
+    if os.getenv('DOMAIN_NAME'):
+        # Pop STAGE value off the request if we have a custom domain
+        here = '/'.join([""]+here.split('/')[2:]) if here.startswith('/{}/'.format(STAGE)) else here
```
kbeamnsidc: I'm probably not looking at this right...but here won't be defined right? Won't the ternary expression fail with 'here' not defined?

PR #108, opened 2020-01-30
Updated if statement to check for 'usr-user-id'

mckadesorensen:


lambda/app.py
```diff
@@ -208,7 +208,7 @@ def logout():
     cookievars = get_cookie_vars(app.current_request.headers)
     template_vars = {'title': 'Logged Out', 'URS_URL': get_urs_url(app.current_request.context)}
 
-    if cookievars:
+    if 'usr-user-id' in cookievars:
```
bbuechler: Can we assume `urs-access-token` is there if `urs-user-id` is?

lambda/app.py
```diff
@@ -80,7 +80,7 @@ def make_redriect(to_url, headers=None, status_code=301):
     headers['Location'] = to_url
     log.debug('to_url: {}'.format(to_url))
     log.debug('headers: {}'.format(headers))
-    return Response(body='', headers=headers, status_code=status_code)
+    return Response(body='', headers=headers, status_code=status_code)ƒ
```
bbuechler: "ƒ"?

lambda/app.py
```diff
@@ -9,7 +9,7 @@
 from rain_api_core.aws_util import get_yaml_file, get_role_session, get_role_creds, check_in_region_request
 from rain_api_core.view_util import get_html_body, get_cookie_vars, make_set_cookie_headers
 from rain_api_core.session_util import get_session, delete_session
-from rain_api_core.egress_util import get_presigned_url, process_varargs, check_private_bucket, check_public_bucket
+from rain_api_core.x_util import get_presigned_url, process_varargs, check_private_bucket, check_public_bucket
```
bbuechler: Is this intentional?

lambda/app.py
mckadesorensen: That's a great question... That escaped my eye.

lambda/app.py
mckadesorensen: Wow, I should have reviewed this...

lambda/app.py
benbart: It would be a very strange situation to have one without the other, since they're set and deleted together. An end user would have had to go and delete one but not the other.

mckadesorensen: I can do an and statement and include that too just incase.

bbuechler: OH MAN, that was close. WAs about to approve when I noticed `usr-user-id`  vs `urs-user-id`

mckadesorensen: Good eye!

PR #116, opened 2020-02-06
Fix lambda updates

bbuechler:
Hotfix for creating build-stamped lambda zips for Terraform

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -1,5 +1,7 @@
 locals {
   vpc_security_group_ids_set = length(var.vpc_security_group_ids) > 0
+  lambda_source_filename     = "${path.module}/lambda.zip"
+  dependency_layer_filename  ="${path.module}/dependencylayer.zip"
 }
 
 resource "aws_security_group" "egress_lambda" {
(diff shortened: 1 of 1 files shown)


lambda/app.py
```diff
@@ -208,7 +208,7 @@ def logout():
     cookievars = get_cookie_vars(app.current_request.headers)
     template_vars = {'title': 'Logged Out', 'URS_URL': get_urs_url(app.current_request.context)}
 
-    if cookievars:
+    if 'urs-user-id' and 'urs-access-token' in cookievars:
```
kbeamnsidc: This should be `if 'urs-user-id' in cookievars and 'urs-access-token' in cookievars:`. The code above will always pass if `urs-access-token` is in cookievars, regardless of if `urs-user-id` is.

```
>>> l = ['a', 'b', 'c']
>>>
>>> if 'a' and 'b' in l: print('yay')
...
yay
>>> if 'q' and 'b' in l: print('yay')
...
yay
>>> if 'q' and 'z' in l: print('yay')
...

```

bbuechler: Damnit. That wasn't even supposed to be in there...

PR #120, opened 2020-02-06
Added endpoint that returns the build #

mckadesorensen:


lambda/app.py
```diff
@@ -232,6 +232,11 @@ def login():
     return make_html_response(template_vars, headers, status_code, 'error.html')
 
 
+@app.route('/version')
```
bbuechler: I think you need to add `version` as a API Gateway endpoint like 

[EgressApiResourceProfile](https://github.com/asfadmin/thin-egress-app/blob/devel/cloudformation/thin-egress-app.yaml#L605) => [EgressAPIMethodProfile](https://github.com/asfadmin/thin-egress-app/blob/devel/cloudformation/thin-egress-app.yaml#L744)

PR #168, opened 2020-05-11
Custom header autotest and PR-2391 bugfixes

benbart:


lambda/app.py
```diff
@@ -190,7 +190,7 @@ def try_download_from_bucket(bucket, filename, user_profile, headers: list):
 
         expires_in = 24 * 3600
         redirheaders['Cache-Control'] = 'private, max-age={0}'.format(expires_in - 60)
-        if isinstance(headers, list):
+        if isinstance(headers, dict):
```
bbuechler, line 193: It looooooks like if you trace it back are enough, headers is _is_ a list? 
https://github.com/asfadmin/rain-api-core/blob/f3ec7baab4967c98f4e8fda144e2ec0071d62557/rain_api_core/egress_util.py#L82-L100

benbart: I need to change it there too

PR #169, opened 2020-05-12
Add '/locate' endpoint 

Jkovarik:
This PR  adds '/locate' endpoint, updates cloudformation/stack to provide it. 

In progress, integration tests are needed, but are dependent on non-public test artifacts

### DIFF
--- a/lambda/app.py
+++ b/lambda/app.py
@@ -285,17 +285,13 @@ def version():
 
 @app.route('/locate')
 def locate():
-    prefix = prepend_bucketname('')
     bucket_name = app.current_request.query_params['bucket_name']
-    if bucket_name.startswith(prefix):
-        bucket_name = bucket_name[len(prefix):]
     bucket_map = collapse_bucket_configuration(get_yaml_file(conf_bucket, bucket_map_file, s3_resource)['MAP'])
     search_map = flatdict.FlatDict(bucket_map, delimiter='/')
     matching_paths = [key for key, value in search_map.items() if value == bucket_name]
     if(len(matching_paths) > 0):
         return json.dumps(matching_paths)
     return Response(body=f'No route defined for {bucket_name}',status_code=404,headers={'Content-Type': 'text/plain'})
-
 
 
 def collapse_bucket_configuration(bucket_map):
--- a/lambda/app.py
+++ b/lambda/app.py
@@ -11,7 +11,7 @@ from rain_api_core.general_util import get_log
 from rain_api_core.urs_util import get_urs_url, do_login, user_in_group
 from rain_api_core.aws_util import get_yaml_file, get_s3_resource, get_role_session, get_role_creds, check_in_region_request
 from rain_api_core.view_util import get_html_body, get_cookie_vars, make_set_cookie_headers_jwt, JWT_COOKIE_NAME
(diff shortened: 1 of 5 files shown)


lambda/requirements.txt
```diff
@@ -1,12 +1,12 @@
-jinja2==2.11.2
-PyYAML==5.3.1
 aws-requests-auth==0.4.2
-pyjwt==1.7.1
+cfnresponse==1.0.2
 chalice==1.14.0
+cryptography==2.9.2
+flatdict==4.0.1
```
Jkovarik, line 5: The real change here is the addition of this in the requirements.     Alpha sorted it for transparency/organization.

lambda/app.py
```diff
@@ -281,6 +283,16 @@ def version():
     return json.dumps({'version_id': '<BUILD_ID>'})
 
 
+@app.route('/locate')
+def locate():
+    search_field = app.current_request.query_params['bucket_name']
```
bbuechler, line 288: is it okay that the bucket_name parameter will need to be the un-prefixed bucket name? If TEA is deployed with a value of `ngap.daac.us-west-2-`,  and you're trying to find bucket  `ngap.daac.us-west-2.dataset-browse-images`, you'll need to supply use `dataset-browse-images` here.

Jkovarik: Good catch.   I'd say for the general use case.... "no". I hadn't considered that given our current use case.

Currently Core is deploying TEA with an enforced lack of a prefix (https://github.com/nasa/cumulus/blob/master/tf-modules/distribution/main.tf#L39)  

As such the implicit expectation is that there isn't one, but that's not useful for TEA users generally.     

It should be easy enough to add the prefix to the returned value if the env var is set.

Jkovarik: cf287c1

This will handle either so long as the user doesn't have a bucket map that's something like 

mapping: <prefix>*

We could error if the query string *does not* have the prefix.   Thoughts?

Jkovarik: Per slack conversation I pulled the prefix filter in e337b00

lambda/app.py
```diff
@@ -282,6 +284,28 @@ def version():
...
 
+@app.route('/locate')
+def locate():
+    bucket_name = app.current_request.query_params['bucket_name']
+    bucket_map = collapse_bucket_configuration(get_yaml_file(conf_bucket, bucket_map_file, s3_resource)['MAP'])
+    search_map = flatdict.FlatDict(bucket_map, delimiter='/')
+    matching_paths = [key for key, value in search_map.items() if value == bucket_name]
+    if(len(matching_paths) > 0):
+        return json.dumps(matching_paths)
+    return Response(body=f'No route defined for {bucket_name}',status_code=404,headers={'Content-Type': 'text/plain'})
```
benbart, line 295: If there is no `bucket_name` param sent, the endpoint will 500 error. Can we make it return an appropriate 400-type error in that case?

benbart: Now that I look at my comment, I think I wasn't being clear. I was using the _royal_ "we".

Jkovarik: No worries, comment was clear.  

5518262 should sort it.  Also added response type, and shortened lines to match local linting.

PR #182, opened 2020-06-25
Bucket region caching and other minor changes.

benbart:


cloudformation/thin-egress-app.yaml
```diff
@@ -101,7 +101,7 @@ Parameters:
 
   LambdaCodeS3Bucket:
     Type: String
-    Default: "asf.public.code"
+    Default: "asf.public.code.usw2"
```
bbuechler, line 104: I don’t think we want this. Public code should all be in one bucket, even if it’s not the “primary” region.

benbart: Isn't this where we're uploading our code to now?

bbuechler: Not that I’m aware of. I think there is a UW2 private code bucket, but public should stay in UE1

benbart: turns out it is:
https://jenkins.asf.alaska.edu/job/rain/job/thin-egress-app-pipeline-prod/79/console
`aws s3 cp ./tea-code-build.79.zip s3://asf.rain.code.usw2/thin-egress-app/`

bbuechler: asf.public.code Vs asf.rain.code

benbart: OK, it's gone

PR #193, opened 2020-08-05
Df.upgrade tea

dpflowersjr:
Removed  public_buckets and private_buckets related files


cloudformation/thin-egress-app.yaml
```diff
@@ -536,8 +536,6 @@ Resources:
           CONFIG_BUCKET: !Ref ConfigBucket
           BUCKET_MAP_FILE: !Ref BucketMapFile
           USE_REVERSE_BUCKET_MAP: !Ref UseReverseBucketMap
-          PUBLIC_BUCKETS_FILE: !Ref PublicBucketsFile
```
bbuechler, line 539: We can remove the Parameters all together.

bbuechler: Which also means removing the params from the Terraform modules also

lambda/app.py
```diff
@@ -23,9 +23,7 @@
 b_map = None
 b_region_map = {}
 bc_client_cache = {}
-public_buckets_file = os.getenv('PUBLIC_BUCKETS_FILE', None)
 public_buckets = None
```
bbuechler, line 26: I think we should also remove any trace of the variables `public_buckets` and `private_buckets` since are no longer needed, right?

Doing so would mean we'd need to update `check_private_bucket()` and `check_public_bucket()` to remove the variables from those definitions.

PR #198, opened 2020-08-13
Token sharing implemented

benbart:
Also handling cases where user doesn't sign the EULA


lambda/requirements.txt
```diff
@@ -10,3 +10,4 @@ pyjwt==1.7.1
 pyOpenSSL==19.1.0 # maybe not necessary
 python-jose==3.2.0
 PyYAML==5.3.1
+deprecated
```
bbuechler, line 13: What is this?

PR #200, opened 2020-08-18
[PR-2488] Add shared token usage documentation

bbuechler:


lambda/app.py
```diff
@@ -577,7 +577,7 @@ def dynamic_url():
         log.debug("Accessing public bucket {0}".format(path))
     elif not user_profile:
         if 'Authorization' in app.current_request.headers and app.current_request.headers['Authorization'].split()[0].lower() == 'bearer':
```
benbart, line 579: I'm a bit concerned about the capitalization here. I'd like to do a check in a case-insensitive manner, but every way I could think of was so unwieldy.

PR #203, opened 2020-08-19
PR-2487 using "app-B" token to fetch profile

benbart:
Also cleaned up the conditionals / excepts for fetching user ID


lambda/app.py
```diff
@@ -94,33 +94,28 @@ def get_user_from_token(token):
...
+            errtxt = f''
+        if 'error_description' in msg:
+            errtxt = errtxt + ' ' + msg['error_description']
 
         log.error(f'Error getting URS userid from token: {errtxt} with code {response.code}')
         log.debug(f'url: {url}, params: {params}, ')
-        return ''
-
-
+    return ''
```
bbuechler, line 118: Can we change this to `None`?  I think the affect will be the same, but will look less confusing!

benbart: I'm returning empty string because I want a string to be returned no matter the result. But I can change it to None.

PR #210, opened 2020-08-24
PR-2507, PR-2508 - Refactor build script to use unittest framework, update build to use k8s agent

dpflowersjr:


build/test/download_test.py
```diff
@@ -0,0 +1,140 @@
...
+        self.assertEqual(r.content, b'["SA/OCN", "SA/OCN_CH", "SB/OCN_CN", "SB/OCN_CH"]')
+
+
+def main(out=sys.stderr, verbosity=2):
+    loader = unittest.TestLoader()
+
+    suite = loader.loadTestsFromModule(sys.modules[__name__])
+    unittest.TextTestRunner(out, verbosity=verbosity).run(suite)
+    s3 = boto3.resource('s3')
+    s3.meta.client.upload_file('/tmp/testresults.json', 'asf.public.code', 'thin-egress-app/testresults.json')
```
bbuechler, line 134: What are we righting out here? This should be a specific JSON object.

build/test/download_test.py
```diff
@@ -0,0 +1,140 @@
+import sys
+import unittest
+import os
+import boto3
+import requests
+from requests.auth import HTTPBasicAuth
+import logging
+
+logging.basicConfig(level=logging.WARNING)
```
bbuechler, line 9: We should probably be doing debug logging, but since we're using boto, we'll want to shut up the noisy output like this: https://github.com/asfadmin/rain-api-core/blob/master/rain_api_core/general_util.py#L24-L30

build/test/download_test.py
```diff
@@ -0,0 +1,281 @@
...
+        final_request = session.get(url, cookies=cookiejar)
+
+        log.info(f"Final request returned: {final_request.status_code} (Expect 200)")
+        self.assertTrue(final_request.status_code == 200)
+
+class authed_download_test(unittest.TestCase):
+    # Check that we get a URS auth redirect for auth'd downloads
+    def test_urs_auth_redirect_for_auth_downloads(self):
+        url = f"{APIROOT}/{METADATA_FILE}"
+        global cookiejar
```
benbart, line 131: would it make sense to have all these auth'd download tests in one class and the `cookiejar` be a member variable?

bbuechler: I thought about doing that, but not until after I'd split it into 3 classes. And now I don't think we can.
There are now 3 classes: **`unauthed`**, **`auth`**, **`authed`**... 
The cookiejar is populated by the **`auth`**, which has only 1 test. Then the cookies is used in all the **`authed`** class cookies.

build/Jenkinsfile
```diff
@@ -21,12 +21,14 @@ pipeline {
     CODE_BUCKET_SUFFIX="""${sh(script:'if [ "${AWS_DEFAULT_REGION}" = "us-west-2" ]; then printf %s ".usw2"; else printf %s ""; fi', returnStdout: true)}"""
   }
 
-  // Build on a slave with docker (for pre-req?)
-  agent { label 'docker' }
+  // Build on a slave with docker on kubernetes
+  agent { kubernetes {
```
bbuechler, line 25: I think there s a missing brace here somewhere? I see 2 opens, 1 closed.

bbuechler:
I suspect we're going to need seperate the K8s agent out or this step:
https://github.com/asfadmin/thin-egress-app/blob/abd44a68e514b196f0f5a18c4d1fe43ba7aab244/build/Jenkinsfile#L85
into a different agent that uses `amazonlinux:2` instead of our custom docker container (docker-registry.asf.alaska.edu:5000/i_and_a_builder_agent:latest) 

We can try this and see if it works though... But it may suffer from the docker-in-docker problem.

PR #213, opened 2020-08-27
Add a semi-function test for bearer token support

bbuechler:
This requires functionality NOT in production EDL.


dpflowersjr:
LGTM

Issue #204, opened 2020-08-20
Consider NOT throwing 500, it gets hijacked by the CloudFront 50X error handling. 


bbuechler:
https://github.com/asfadmin/thin-egress-app/blob/eb85b1498dbf36ca0adf3b248207d6b04093490d/lambda/app.py#L239-L243

Its not a HUGE deal and if you have a CloudFront error configured you see that. But it does mask the error a bit. You see this:
![image](https://user-images.githubusercontent.com/1580924/90708379-19ffbd00-e246-11ea-9688-bd9e743f342e.png)
instead of the actual error.

PR #225, opened 2020-08-31
#204 Passing error code from get_bucket_region() should one happen

benbart:


lambda/app.py
```diff
@@ -408,19 +413,19 @@ def get_range_header_val():
...
     params = {}
     now = time.time()
-    
+
     if user_id not in bc_client_cache:
         # This a new user, generate a new bc_Config client
         params['config'] = bc_Config(**get_bcconfig(user_id))
         session = get_role_session(user_id=user_id)
         bc_client_cache[user_id] = {"client": session.client('s3', **params), "timestamp": now }
-    elif now - bc_client_cache[user_id]["timestamp"] >= (50*60):
```
bbuechler, line 417: How'd these spaces sneak in here?!??! I thought I was being so careful!

PR #243, opened 2020-09-24
Update and expand README

bbuechler:


README.MD
```diff
@@ -59,7 +205,7 @@ DEPENDENCYLAYERFILENAME=tea_depencency_layer.zip
 git clone https://github.com/asfadmin/thin-egress-app
 cd thin-egress-app
 
-# copy requirements to a directory that 
+# copy requirements to a directory that is mounted into the docker container
```
benbart, line 208: I think the depbuild script is changed enough these days that we'll have to redo this section. I can do that after I take a look at the current state of things

Issue #260, opened 2020-10-15
Make Lambda memory allocation dynamic, default to 1,792 MB


bbuechler:
We seen testing at scale that normal usage is closer to 400MB, but spikes to near 700MB. However, at 1792 MB, we can allocate an entire vCPU to help mitigate backgrounding. 

1792MB seems a little on the excessive side when normal usage is around 400 and spikes only go up to 700, however Lambda only charges you for RAM used, not RAM allocated. Getting that full proportional vCPU should offset any concerns of wasted RAM.

We should add a new Parameter to the CF & TF to support dynamic setting of the Lambda Memory allocation, and have it default to 1792. 

https://docs.aws.amazon.com/lambda/latest/dg/configuration-console.html#:~:text=AT%201%2C792%20MB%2C%20a%20function%20has%20the%20equivalent%20of%20one%20full%20vCPU

benbart:
This was implemented with PR #261

PR #261, opened 2020-10-16
#260 Made adjustable memory amount. Default of 1792

benbart:


bbuechler:
We need to ALSO update the TF stuff...  Same story, with default.

bbuechler:
I can't remember how CloudFormation handles _new_ default params, we may need to add it to https://github.com/asfadmin/thin-egress-app/blob/bbarton-issues-202010/build/deploy.sh#L284 and https://github.com/asfadmin/thin-egress-app/blob/bbarton-issues-202010/build/Jenkinsfile#L196

BUT we should definitely add it here: https://github.com/asfadmin/thin-egress-app/blob/bbarton-issues-202010/README.MD#all-parameters probably here: https://github.com/asfadmin/thin-egress-app/blob/bbarton-issues-202010/README.MD#using-awscli

bbuechler:
awesome.

PR #266, opened 2020-10-23
Git submodule that supports new logging features

### LINKED ISSUE #258: Stamp each request with a unique-ish ID (UUID) to make searching logs easier
Every [`@app.route()`](https://github.com/asfadmin/thin-egress-app/blob/devel/lambda/app.py#L550) should generate a unique id (perhaps a UUID or _part_ of a UUID. That ID would be appended to **ALL** log entries and would allow easier filtering in logs. 

This _could_ be done by adding a global variable "REQ_ID" and adding that the log handler in the way we stamp [build id](https://github.com/asfadmin/rain-api-core/blob/64fa3fe35534e1971ae3bf19c2493157f0b6e613/rain_api_core/general_util.py#L8)

benbart:
Also minor adjustments to jenkinsfile

closes #258


dpflowersjr:
LGTM

PR #287, opened 2020-11-20
Bugfixes

### LINKED ISSUE #284: Catch error and return meaningful client error when Secret is not found
https://github.com/asfadmin/thin-egress-app/blob/18686022c7e67c62b557441defad0fcd72009cf8/lambda/app.py#L372

During the log in process, `rain-api-core` will [re-raise](https://github.com/asfadmin/rain-api-core/blob/master/rain_api_core/aws_util.py#L66) an `ClientError` error if a secret is not found. This behavior in `rain-api-core` is desirable, however TEA should try to catch the `ClientError` and provide a meaningful error outward to facilitate easier troubleshooting.

### LINKED ISSUE #285: Don't fail if inbound requests has no headers
In normal operation, there is  **always** at least one header that arrives with the lambda event payload. 

However, when you run a API Gateway Method Execution Test, unless you explicitly supply a header key pair, no headers actually end up in the event payload and `event['headers']` is `None`. That causes THIS line:

https://github.com/asfadmin/thin-egress-app/blob/18686022c7e67c62b557441defad0fcd72009cf8/lambda/app.py#L46

... to fail with this error:

```
File "/var/task/app.py", line 46, in __call__
userid = get_jwt_field(get_cookie_vars(event['headers']), 'urs-user-id')
  File "/var/task/rain_api_core/view_util.py", line 101, in get_cookie_vars
cooks = get_cookies(headers)
  File "/var/task/rain_api_core/view_util.py", line 129, in get_cookies
c = hdrs.get('cookie', hdrs.get('Cookie', hdrs.get('COOKIE', None)))
```

...and then the result in the Test harness is:

**Request**: /version
**Status**: 502
**Latency**: 1314 ms
**Response Body**:
```
{
  "message": "Internal server error"
}
```
**Response Headers**
```
{"x-amzn-ErrorType":"InternalServerErrorException"}
```

I think fixings this requires that we find all references to `event['headers']` and replace with something like `event['headers'] or {}`

benbart:
closes #285 
closes #284

### DIFF
--- a/lambda/app.py
+++ b/lambda/app.py
@@ -369,7 +369,16 @@ def logout():
 
 @app.route('/login')
 def login():
-    status_code, template_vars, headers = do_login(app.current_request.query_params, app.current_request.context, os.getenv('COOKIE_DOMAIN', ''))
+    try:
+        status_code, template_vars, headers = do_login(app.current_request.query_params, app.current_request.context, os.getenv('COOKIE_DOMAIN', ''))
+    except ClientError as e:
+        log.error(e)
+        status_code = 500
+        headers = {'x-request-id': app.lambda_context.aws_request_id}
+        template_vars = {
+            'contentstring': 'Client Error occurred. ',
+            'title': 'Client Error',
+        }
     if status_code == 301:
         return Response(body='', status_code=status_code, headers=headers)
 
--- a/lambda/app.py
+++ b/lambda/app.py
@@ -43,7 +43,7 @@ class TeaChalice(Chalice):
         log_context(route=resource_path, request_id=context.aws_request_id)
         # get_jwt_field() below generates log messages, so the above log_context() sets the
         # vars for it to use while it's doing the username lookup
-        userid = get_jwt_field(get_cookie_vars(event['headers']), 'urs-user-id')
+        userid = get_jwt_field(get_cookie_vars(event.get('headers', {})), 'urs-user-id')
         log_context(user_id=userid)
 
         resp = super().__call__(event, context)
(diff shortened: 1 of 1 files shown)


lambda/app.py
```diff
@@ -369,7 +369,16 @@ def logout():
...
 @app.route('/login')
 def login():
-    status_code, template_vars, headers = do_login(app.current_request.query_params, app.current_request.context, os.getenv('COOKIE_DOMAIN', ''))
+    try:
+        status_code, template_vars, headers = do_login(app.current_request.query_params, app.current_request.context, os.getenv('COOKIE_DOMAIN', ''))
+    except ClientError as e:
+        log.error(e)
+        status_code = 500
+        headers = {'x-request-id': app.lambda_context.aws_request_id}
+        template_vars = {
```
benbart, line 378: Do we want to provide the aws request id to user-facing web page?

lambda/app.py
```diff
@@ -43,7 +43,7 @@ def __call__(self, event, context):
         log_context(route=resource_path, request_id=context.aws_request_id)
         # get_jwt_field() below generates log messages, so the above log_context() sets the
         # vars for it to use while it's doing the username lookup
-        userid = get_jwt_field(get_cookie_vars(event['headers']), 'urs-user-id')
+        userid = get_jwt_field(get_cookie_vars(event.get('headers', {})), 'urs-user-id')
```
bbuechler, line 46: I tried this approach this AM and since `headers` is defined as `None`, you still get none. 
```
>>> event = {'headers': None}
>>> print (event.get('headers', {}))
None
>>> print ( event['headers'] )
None
>>> print ( event['headers'] or {})
{}
```

benbart: You already fixed this?

bbuechler: No, when I was writing the ticket, I was curious if `event.get('headers', {})` would do the trick. My testing showed it would not.

lambda/app.py
bbuechler: yes!

dpflowersjr:
LGTM

PR #295, opened 2020-12-03
Df.upgrade tea

dpflowersjr:


lambda/requirements.txt
```diff
@@ -7,8 +7,9 @@ jinja2==2.11.2
 jwcrypto==0.8
 netaddr==0.8.0
 pyjwt==1.7.1
-pyOpenSSL==20.0.0 # maybe not necessary
+pyOpenSSL==19.1.0 # maybe not necessary
 python-jose==3.2.0
 PyYAML==5.3.1
+ipaddress==1.0.23
```
bbuechler, line 13: Where do you see this resolves the issue? WhiteSource seems to suggest `ipaddress-1.0.23-py2.py3-none-any.whl` is still an issue. 
image

lambda/requirements.txt
```diff
@@ -7,8 +7,9 @@ jinja2==2.11.2
 jwcrypto==0.8
 netaddr==0.8.0
 pyjwt==1.7.1
-pyOpenSSL==20.0.0 # maybe not necessary
+pyOpenSSL==19.1.0 # maybe not necessary
```
bbuechler, line 10: What vulnerability is this address? I don't see it in the issues

Issue #328, opened 2021-03-05
Create mechanism to flush Lambda env and reload config


bbuechler:
Need a mechanism to pull fresh config TEA on demand. This should be able to function independent of an application re-deployments. 

Potentially utilize a hot-updatable Lambda env variable to trigger a soft-flush of environment. 

How do we want to trigger the flush? AWS API level Invocation of a VPC-restricted? Auth'd 😬 API Endpoint?

bbuechler:
```bash
# AWS account stuff:
export tealambda=your-cumulus-dev-thin-egress-app-EgressLambda
# Don't forget to make sure you provide a AWS Region and AWS Profile
# This requires jq is installed.

# Grab the current Lambda environment variables, set OR update the BUMP variable to unix timestamp
current_lambda_env=$(aws lambda  get-function-configuration --function-name $tealambda | jq --arg now `date +%s` '.Environment.Variables.BUMP=$now'  | jq .Environment.Variables )

# Build the update object:
updated_lambda_env=$(echo '{"Variables":'$current_lambda_env'}')

# Update the lambda:
aws lambda update-function-configuration --function-name $tealambda --environment "$updated_lambda_env"
```

bbuechler:
PR#333 and PR#334 Add this functionality!

PR #333, opened 2021-04-22
Creation of TEAbumper lambda #328

benbart:
Also updated the rain-api-core

### DIFF
--- a/cloudformation/thin-egress-app.yaml
+++ b/cloudformation/thin-egress-app.yaml
@@ -947,3 +947,28 @@ Resources:
     Properties:
       ServiceToken: !GetAtt UpdatePolicyLambda.Arn
 
+  BumperLambda:
+    Type: AWS::Lambda::Function
+    DependsOn:
+      - EgressLambdaDependencyLayer
+    Properties:
+      Code:
+        S3Bucket: !Ref LambdaCodeS3Bucket
+        S3Key: !Ref LambdaCodeS3Key
+      Role: !GetAtt EgressLambdaIamRole.Arn
+      FunctionName: !Sub "${AWS::StackName}-BumperLambda"
+      VpcConfig:
+        !If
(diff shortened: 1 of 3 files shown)


cloudformation/thin-egress-app.yaml
```diff
@@ -947,3 +947,59 @@ Resources:
...
+  BumperLambda:
+    Type: AWS::Lambda::Function
+    Properties:
+      Code:
+        S3Bucket: !Ref LambdaCodeS3Bucket
+        S3Key: !Ref LambdaCodeS3Key
+      Role: !GetAtt BumperLambdaIamRole.Arn
+      FunctionName: !Sub "${AWS::StackName}-BumperLambda"
+      VpcConfig:
+        !If
```
bbuechler, line 959: I honestly have no idea... If the function is in the VPC, can you call it on the cmd line from outside the vpc?

lambda/tea_bumper.py
```diff
@@ -0,0 +1,31 @@
...
+
+    log.debug(f"envvar for {TEA_LAMBDA_NAME}: {egress_env['Variables']}")
+    response = client.update_function_configuration(
+        FunctionName=TEA_LAMBDA_NAME,
+        Environment=egress_env
+    )
+    log.debug(response)
+
+
+def version():
```
bbuechler, line 30: Is this to make logging work?

cloudformation/thin-egress-app.yaml
benbart: I just copied it from the egress lambda. My understanding of networking is thin enough that I didn't want to mess with something I didn't understand. If I should remove it, I'd be happy to.

lambda/tea_bumper.py
benbart: It's to provide a hardcoded documentation of the version in file

PR #349, opened 2021-05-05
Fixed mistake that included sensitive information

mckadesorensen:


bbuechler:
Careful. This is a PR to master

PR #354, opened 2021-05-06
Added JWT Blacklist test to download_test.py

mckadesorensen:


build/test/download_test.py
```diff
@@ -0,0 +1,432 @@
...
+
+            aws_lambda_client = boto3.client('lambda')
+            aws_function_name = f'{STACKNAME}-EgressLambda'
+
+            lambda_configuration = aws_lambda_client.get_function_configuration(
+                FunctionName=aws_function_name
+            )
+
+            new_env_vars = lambda_configuration["Environment"]
+            new_env_vars["Variables"].update(endpoint_dict)
```
bbuechler, line 362: A lot of this is duplicated in between these two new tests. Perhaps some of the duplication can be moved into a helper function?

bbuechler:
This is a PR to `master` again.

PR #357, opened 2021-05-06
Added JWT test

mckadesorensen:


build/test/download_test.py
```diff
@@ -299,13 +299,84 @@ def test_vallidate_bearer_token_works(self):
...
+        log.info(f"Update status: {env_vars_update}")
+
+        return lambda_configuration
+
+    def test_validate_invalid_jwt(self):
+        url = f"{APIROOT}/{METADATA_FILE}"
+        global cookiejar
+
+        try:
+            aws_lambda_client = boto3.client('lambda')
```
benbart, line 331: I would suggest making this a member variable in the class.

build/test/download_test.py
```diff
@@ -299,13 +299,84 @@ def test_vallidate_bearer_token_works(self):
...
+
+            r = requests.get(url, cookies=cookiejar, allow_redirects=False)
+            log.info(f"Blacklisted JWTs should result in a redirect to EDL. r.is_redirect: {r.is_redirect} (Expect True)")
+            self.assertTrue(r.is_redirect)
+        except Exception as e:
+            log.info(e)
+            self.assertTrue(False)
+
+        log.info("Reverting to original environment variables")
+        self.set_original_env_vars(aws_lambda_client, aws_function_name, lambda_configuration["Environment"])
```
benbart, line 347: If the code above throws an exception before one of these is set, there will be problems.

build/test/download_test.py
benbart: Maybe a lot of the setup below could be done in the constructor. Or isn't there some kind of set up and tear down built into the unit test parent class?

PR #415, opened 2021-10-13
Correct spelling of region

abarciauskas-bgse:


bbuechler:
Thanks. I'm a terrible speller, and an almost worst typer, so keep em' coming. Theres LOTS more peppered throughout.

PR #431, opened 2021-11-22
Update TEA Lambda to attempt to reduce CIDR range

ccuadrado:
This change came about because the "DownloadRoleInRegion" lambda update was failing to properly update, complaining that the role size was too big.

We were deploying in us-east-1, so the CIDR range fetched was about 799 addresses.
With cidr_merge it got it down to about 449, or saving about 6000 bytes on the role.

This should allow the role policy document to be adequately created for now. It will also dynamically adjust as AWS changes their IP list.


ccuadrado:
Resubmitting branched from devel instead of master.

PR #438, opened 2022-01-04
rename lambda folder

Jlrine2:
renames the lamdba folder to src to prevent python nameing collisions and preserves history for any moved files


Askaholic:
Probably also need to update this line in the .gitignore:
```
/lambda/.chalice
```

reweeden:
IMO naming it `tea` or `thin_egress_app` would make more sense as it will need to be imported as a python module for the unit tests.

PR #440, opened 2022-01-08
PR-2930 add unit tests

reweeden:
For the most part I just wrote tests that document how the functions currently behave, although there were a few places where I did find some bugs that I went ahead and fixed. I also did some light refactoring of really small stuff and obvious pep8 violations. Afterwards I also refactored a few other things including replacing the hand rolled caching with a proper library. These are split out into their own commits so they can be moved to a different PR if needed.


lambda/app.py
```diff
@@ -3,39 +3,48 @@
...
 bucket_map_file = os.getenv('BUCKET_MAP_FILE', 'bucket_map.yaml')
 b_map = None
-b_region_map = {}
-bc_client_cache = {}
-s3_resource = get_s3_resource()
+# TODO(reweeden): Refactor when wrapped attributes are implemented
+# https://github.com/tkem/cachetools/issues/176
+get_bucket_region_cache = cachetools.LRUCache(maxsize=128)
+# TODO(reweeden): Why is this even here? This variable is unused!
+# s3_resource = get_s3_resource()
```
reweeden, line 45: Can this be deleted?

lambda/app.py
```diff
@@ -73,17 +82,29 @@ def __init__(self, payload: dict):
 
 
 def get_request_id() -> str:
+    assert app.lambda_context is not None
+
     return app.lambda_context.aws_request_id
 
+
+# TODO(reweeden): fix typing
+# typing module causes errors in AWS lambda?
```
reweeden, line 91: I noticed a comment in the dependency builder that there were some errors with the `typing` module. What is the error and does that mean we can't use the `typing` module at all?

lambda/app.py
benbart: Likely an abandoned road. It can be deleted.

lambda/app.py
```diff
@@ -431,13 +463,14 @@ def logout():
 @app.route('/login')
 def login():
     try:
+        headers = {}
         aux_headers = get_aux_request_headers()
         status_code, template_vars, headers = do_login(app.current_request.query_params, app.current_request.context,
                                                        os.getenv('COOKIE_DOMAIN', ''), aux_headers=aux_headers)
     except ClientError as e:
-        log.error(e)
+        log.error("%s", e)
```
benbart, line 471: Why not fix the bug in the rain api core?

reweeden: I should do that as well

requirements.txt
```diff
@@ -1,8 +1,10 @@
 aws-requests-auth==0.4.3
+cachetools==5.0.0
 cfnresponse==1.1.1
 chalice==1.26.2
 cryptography==35.0.0
 flatdict==4.0.1
+git+https://github.com/asfadmin/rain-api-core.git@c9b9f0ed07ab99d44c32c1caec3684388091d640
```
reweeden, line 7: Will need to make sure that templates can still be fetched. Right now rain-api-core always computes the template directory relative to the location of `view_util.py` which may be in a different location when installed through pip. This should be refactored to a config variable.

lambda/tea_bumper.py
```diff
@@ -29,5 +28,6 @@ def lambda_handler(event, context):
     log.debug(response)
 
 
+# TODO(reweeden): This appears to be unused
 def version():
     return json.dumps({'version_id': '<BUILD_ID>'})
```
reweeden, line 33: Anyone know if this function is actually used somehow? It's not called anywhere within the codebase.

requirements.txt
reweeden: Addressed in https://github.com/asfadmin/rain-api-core/pull/165

lambda/app.py
```diff
@@ -456,13 +463,14 @@ def logout():
...
         status_code = 500
-        headers = {'x-request-id': app.lambda_context.aws_request_id}
+        # headers = {'x-request-id': app.lambda_context.aws_request_id}
```
reweeden, line 473: Delete?

reweeden: Seems like chalice includes this header automatically

lambda/tea_bumper.py
benbart: It isn't used. It can be removed.

.gitignore
```diff
@@ -1,13 +1,25 @@
...
 .DS_Store
-/lambda/.chalice
-/.idea
+
+# Tools
 .terraform
-/venv
-/pkg
 /terraform/providers.tf
-bucket_map.yaml
```
Jlrine2, line 9: any particular reason these were removed?

lambda/tea_bumper.py
```diff
@@ -12,6 +11,7 @@
 
 
 def lambda_handler(event, context):
+    del event
```
Jlrine2, line 14: I get that it isn't used, but is it necessary to delete?

tests/__init__.py
```diff
@@ -0,0 +1,8 @@
+import pathlib
+from typing import IO, Any
+
+RESOURCES_PATH = pathlib.Path(__file__).parent.joinpath("resources/").absolute()
+
+
+def open_resource(file: str, *args, **kwargs) -> IO[Any]:
```
Jlrine2, line 7: only used in conftest, would it be better to define it their?

Jlrine2 (approved):
Looks good, couple of non important style suggestions but I think it is good to merge.

lambda/tea_bumper.py
benbart: I wouldn't say "necessary", but it's never going to be used for anything going forward. It's a leftover from early design when the tea_bumper was to be an http endpoint.

reweeden: This is just what the openstack style guide says to do with unused arguments, so I've been trying it out. It not only documents explicitly that a certain argument is unused but it also makes it impossible for someone to go in later and start using the argument without updating the documentation (you will get 'undefined name' error unless you remove the del statement). The other options are:

1. Do nothing - Not a good idea IMO because sooner or later you will end up with unused parameters bloating your function signatures that really should be deleted but nobody noticed.
2. Rename the variable to `_` or `_event` - This is what I have done in the past, the only downside is that it doesn't always work because it breaks the function API wrt keyword arguments, as now the parameter has a different name.

.gitignore
reweeden: They weren't removed, I just grouped them a bit to make things more organized

tests/__init__.py
reweeden: Yea probably

reweeden: Although the conftest file is more for pytest configuration i.e. the fixture that wraps this function. If you were to just import it directly I feel like it makes more sense to do `from tests import open_resource` than `from tests.conftest import open_resource`. But maybe it should just be a 'private' function anyways

reweeden: I looked for pytest plugins that provide this functionality, but I couldn't find any that worked in a way that I liked.

tests/test_app.py
```diff
@@ -0,0 +1,1190 @@
...
+    response = app.try_download_from_bucket("somebucket", "somefile", None, {})
+    assert response.body == "Invalid Range"
+    assert response.status_code == 416
+    assert response.headers == {}
+    mock_get_role_creds.assert_called_once()
+    mock_get_role_session.assert_called_once()
+    mock_check_in_region_request.assert_called_once()
+
+
+@mock.patch("lambda.app.JWT_COOKIE_NAME", "asf-cookie")
```
reweeden, line 482: MODULE format string

tests/test_app.py
```diff
@@ -0,0 +1,1190 @@
...
+        },
+        {"Content-Type": "text/html"},
+        200,
+        "root.html"
+    )
+
+
+@mock.patch(f"{MODULE}.get_urs_url", autospec=True)
+@mock.patch(f"{MODULE}.get_cookie_vars", autospec=True)
+@mock.patch("lambda.app.JWT_COOKIE_NAME", "asf-cookie")
```
reweeden, line 509: Same here

tests/test_app.py
```diff
@@ -0,0 +1,1190 @@
...
+
+def test_profile(client):
+    response = client.http.get("/profile")
+
+    assert response.body == b"Profile not available."
+    assert response.status_code == 200
+
+
+@mock.patch(f"{MODULE}.get_jwt_keys", autospec=True)
+@mock.patch("lambda.app.JWT_ALGO", "THE_ALGO")
```
reweeden, line 1176: And here

tests/test_update_lambda.py
```diff
@@ -0,0 +1,144 @@
+import importlib
+import io
+import json
+from unittest import mock
+
+import pytest
+
+MODULE_PATH = "lambda.update_lambda"
```
reweeden, line 8: And here

PR #441, opened 2022-01-10
Add GitHub Actions for lints and tests

reweeden:
isort and flake8 can (and should) be configured using a `setup.cfg`. Still need to see about a coverage reporting tool.

Once we get this working how we like we can refactor to a reusable workflow.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,23 @@
+name: lint
+on:
+  pull_request:
+
+jobs:
+  isort:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - run: pip install -r lambda/requirements.txt
+      - uses: isort/isort-action@master
+
+  flake8:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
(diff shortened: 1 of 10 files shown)


.github/workflows/lint.yml
```diff
@@ -0,0 +1,18 @@
+name: lint
+on:
+  pull_request:
```
reweeden, line 3: Maybe run on branches too?

.github/workflows/test.yml
```diff
@@ -0,0 +1,28 @@
...
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r lambda/requirements.txt pytest
+
+    - run: pip install pytest-github-actions-annotate-failures
+
+    - run: pytest --cov=lambda --cov-report=term-missing --cov-report=xml --cov-branch tests
+
+    # TODO(reweeden): Report coverage
```
reweeden, line 28: Codecov is free for open source. We could use it on this repo.

mckadesorensen: That sounds good to me.

.github/workflows/test.yml
```diff
@@ -0,0 +1,31 @@
+name: test
+
+on:
+  pull_request:
+  push:
+    branches:
+      - devel
+      - master
```
Jlrine2, line 8: any particular reason not to run on push to any branch?

Jlrine2 (approved):

.github/workflows/test.yml
reweeden: Just so it doesn't run twice on PRs.

PR #443, opened 2022-01-24
PR-3041 Update dependencies

reweeden:
Might be a good idea to pin dependency versions with `pip-tools` or `pipenv`. These tools will automatically pin the entire dependency tree in order to produce deterministic builds. They also both provide a single command upgrade mechanism to pin everything to the latest version.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,16 +1,8 @@
-aws-requests-auth==0.4.3
 cachetools==5.0.0
 cfnresponse==1.1.1
 chalice==1.26.2
-cryptography==35.0.0
 flatdict==4.0.1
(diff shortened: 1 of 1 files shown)


Jlrine2 (approved):

Jlrine2:
PR looks like it was based of master, requirements are now in `/requirements.txt` and `/requirements-dev.txt`

Jlrine2 (changes requested):
^

reweeden:
> PR looks like it was based of master, requirements are now in `/requirements.txt` and `/requirements-dev.txt`

No it was based on devel. The requirements.txt was relocated in #440 which wasn't merged at the time.

Jlrine2 (approved):

PR #451, opened 2022-02-02
Ltt 61 local deployment update

ccuadrado:
This looks better. Reconfigure deploy.sh script so it can be used to deploy a cloudformation stack in the (optional) EDC for development purposes.

### DIFF
--- a/build/deploy.sh
+++ b/build/deploy.sh
@@ -11,10 +11,12 @@
 #              --region-name="aws-region-value" \
 #              --maturity=<SBX|DEV|SIT|INT|UAT|TEST|PROD>  \
 #              --edl-user-creds='<EDL-USERNAME>:<EDL-PASSWORD>'
+#              --local_build
 #
 #        -a|--aws-profile     AWS Profile (from ~/.aws/credentials)
 #        -b|--bastion         SSM Bastion name ("NGAP SSH Bastion"?)
 #        -c|--client-id       EDL App Client ID
+#        -d|--domain          Custom domain to use on Redirect (not API)
 #        -e|--edl-user-creds  EDL USER credentials (For Validating DL's)
 #        -k|--key-file        ssh key for connecting to SSM Bastion
 #        -m|--maturity        Account/EDL Matuirty (SBX|SIT|UAT|PROD)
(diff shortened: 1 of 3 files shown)


build/deploy.sh
```diff
@@ -248,38 +259,109 @@ for i in $config_bucket $code_bucket $restricted_bucket $public_bucket; do
...
+                                          --output=text | head -1 | xargs -n1 | grep yaml)
+   cf_yaml_name=$(echo $cf_yaml | cut -d "/" -f 2)
+   echo "code zip is: $code_zip"
+   echo "layer_zip is: $layer_zip"
+   echo "cf yaml is: $cf_yaml"
+   echo "cf_yaml_name is: $cf_yaml_name"
+
+else 
+   echo "Building from local source..."
+   # Begin code deployment packaging.
```
reweeden, line 285: BTW I'm working on writing a Makefile that will know how to do all of these things, so hopefully this whole branch can become something like `make build`.

npauzenga:
@reweeden and @ccuadrado is this PR still valid? I see the Makefile changes have been merged. Do we still need these changes? @ccuadrado are you able to use the Makefile in `devel` to produce and deploy a build? I've had success making a Terraform build with it.

reweeden:
Yea, let me know if the `make deploy` targets work for your development purposes. I think the idea of the `deploy.sh` file is to be more for production deployments and `make deploy` for development and testing. I did intentionally leave some stuff out of the makefile though like creating buckets and secrets because it was simpler that way and I didn't want the makefile to be creating a bunch of resources that might be hard to track down and delete later. But if some of these would be really helpful then maybe they are worth putting in after all.

npauzenga:
@reweeden @ccuadrado I think we can go ahead and close this PR. The changes here are really for a user without a Cumulus stack or a previous TEA deployment to be able to build from scratch. If that turns out to be a valuable use case perhaps a version of this could be revived in the future.

In the meantime, the new Makefile works great for someone that has set up TEA at least once and has those assumed resources already created. This deploy script worked to get someone without that infrastructure up and running so I think we can say we got what we needed from it.

reweeden:
Sounds good!

PR #452, opened 2022-02-03
PR-3033 Add a makefile for automating development tasks

reweeden:
I've been writing the Makefile with a focus mostly on development tasks plus producing all build artifacts. It can deploy a development stack minus secrets, and due to the magic of Make will update the relevant portions when you make changes to the code. For instance if you edit the `app.py` and then run `make deploy` it will rebuild the lambda code zip file, upload it to S3 and update the CloudFormation stack. If you then edit the requirements.txt and run `make deploy` again it will rebuild the dependency layer zip file, upload it to S3 and update the CloudFormation stack. This means it takes round about 1 minute from making a local change to having it take effect on your development stack.

TODO: 
- Use Makefile in Jenkins
- Document how to run the unit tests in a virtual environment
- Resolve TODO's ;)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,79 +1,133 @@
 # TODO(reweeden): docs
+
 SOURCES = \
 	lambda/app.py \
 	lambda/tea_bumper.py \
 	lambda/update_lambda.py
 
+# TODO(reweeden): Fix this so changing templates causes a rebuild
 RESOURCES = \
 	lambda/templates
 
-DIST = $(SOURCES:lambda=dist/code)
-DIST_RESOURCES = $(RESOURCES:lambda=dist/code)
+# Output directory
+DIR = dist
+EMPTY = $(DIR)/empty
+# Output artifacts
+DIST = $(SOURCES:lambda=$(DIR)/code)
+DIST_RESOURCES = $(RESOURCES:lambda=$(DIR)/code)
 
-CODE_BUCKET =
-CODE_DIR = tea-dev
-
-AWS_PROFILE = default
-AWS_DEFAULT_REGION = us-west-2
-
-JWTALGO = RS256
-JWTKEYSECRETNAME = bbarton_rsa_keys_4_jwt
-LAMBDA_TIMEOUT = 6
-LAMBDA_MEMORY = 128
-URS_URL = https://uat.urs.earthdata.nasa.gov
-URS_CREDS_SECRET_NAME = URS_creds_ASF_DATA_ACCESS_EGRESS_CONTROL_UAT
-BUCKETNAME_PREFIX = rain-uw2-t-
-
+DATE := $(shell date --utc "+%b %d %Y, %T %Z")
+DATE_SHORT := $(shell date --utc "+%Y%m%dT%H%M%S")
 BUILD_ID := $(shell git rev-parse --short HEAD)
 
-.PHONY: build clean default deploy deploy-dependencies deploy-code deploy-cloudformation layer-builder test
+#####################
+# Deployment Config #
+#####################
+
+# A tag to discriminate between artifacts of the same name. The tag should be
+# different for each build.
+S3_ARTIFACT_TAG = $(DATE_SHORT)
 
 default:
 	@echo "WIP"
 	@echo "BUILD_ID: ${BUILD_ID}"
 
-build: dist/code-$(BUILD_ID).zip dist/dependencies-$(BUILD_ID).zip dist/thin-egress-app-$(BUILD_ID).yaml
+# Include custom configuration
+CONFIG:
+	@echo "It looks like you are building TEA for the first time.\nPlease review the configuration in '$@' and run Make again.\n"
+	@cp --no-clobber CONFIG.example CONFIG
+	@exit 1
 
+include CONFIG
+
+##############################
+# Local building/development #
+##############################
+
+# Build everything
+.PHONY: build
+build: \
+	$(DIR)/thin-egress-app-code.zip \
+	$(DIR)/thin-egress-app-dependencies.zip \
+	$(DIR)/thin-egress-app.yaml
+
+# Build individual components
+.PHONY: dependencies
+dependencies: $(DIR)/thin-egress-app-dependencies.zip
+	@echo "Built dependency layer for version ${BUILD_ID}"
+
+.PHONY: code
+code: $(DIR)/thin-egress-app-code.zip
+	@echo "Built code for version ${BUILD_ID}"
+
+.PHONY: yaml
+yaml: $(DIR)/thin-egress-app.yaml
+	@echo "Built CloudFormation template for version ${BUILD_ID}"
+
+.PHONY: clean
 clean:
-	rm -r dist
+	rm -r $(DIR)
 
-dist/thin-egress-app-$(BUILD_ID).yaml: cloudformation/thin-egress-app.yaml
-	mkdir -p dist
-	cp cloudformation/thin-egress-app.yaml dist/thin-egress-app-$(BUILD_ID).yaml
-	sed -i -e "s/asf.public.code/${CODE_BUCKET}/" dist/thin-egress-app-$(BUILD_ID).yaml
-	sed -i -e "s/<CODE_ARCHIVE_PATH_FILENAME>/${CODE_DIR}\\/code-${BUILD_ID}.zip/" dist/thin-egress-app-$(BUILD_ID).yaml
-	sed -i -e "s/<DEPENDENCY_ARCHIVE_PATH_FILENAME>/${CODE_DIR}\\/dependencies-${BUILD_ID}.zip/" dist/thin-egress-app-$(BUILD_ID).yaml
-	sed -i -e "s/<BUILD_ID>/${BUILD_ID}/g" dist/thin-egress-app-$(BUILD_ID).yaml
-	sed -i -e "s/^Description:.*/Description: \"TEA snapshot, version: ${BUILD_ID}\"/" dist/thin-egress-app-$(BUILD_ID).yaml
+$(DIR)/thin-egress-app-dependencies.zip: requirements.txt | $(DIR)
+	WORKSPACE=`pwd` DEPENDENCYLAYERFILENAME=$(DIR)/thin-egress-app-dependencies.zip build/dependency_builder.sh
 
-dist/code-$(BUILD_ID).zip: $(DIST) $(DIST_RESOURCES)
-	mkdir -p dist/code
-	cp -r $(DIST) dist/code
-	cp -r $(DIST_RESOURCES) dist/code
-	find dist/code -type f -exec sed -i "s/<BUILD_ID>/${BUILD_ID}/g" {} \;
-	cd dist/code && zip -r ../code-$(BUILD_ID).zip .
+$(DIR)/thin-egress-app-code.zip: $(DIST) $(DIST_RESOURCES) | $(DIR)/code
+	cp -r $(DIST) $(DIR)/code
+	cp -r $(DIST_RESOURCES) $(DIR)/code
+	find $(DIR)/code -type f -exec sed -i "s/<BUILD_ID>/${BUILD_ID}/g" {} \;
+	cd $(DIR)/code && zip -r ../thin-egress-app-code.zip .
 
-dist/dependencies-$(BUILD_ID).zip: requirements.txt
-	mkdir -p dist
-	WORKSPACE=`pwd` DEPENDENCYLAYERFILENAME=dist/dependencies-$(BUILD_ID).zip build/dependency_builder.sh
+$(DIR)/thin-egress-app.yaml: cloudformation/thin-egress-app.yaml | $(DIR)
+	cp cloudformation/thin-egress-app.yaml $(DIR)/thin-egress-app.yaml
+	sed -i -e "s/asf.public.code/${CODE_BUCKET}/" $(DIR)/thin-egress-app.yaml
+	# sed -i -e "s/<CODE_ARCHIVE_PATH_FILENAME>/${CODE_DIR}\\/code-${BUILD_ID}.zip/" $(DIR)/thin-egress-app.yaml
+	# sed -i -e "s/<DEPENDENCY_ARCHIVE_PATH_FILENAME>/${CODE_DIR}\\/dependencies-${BUILD_ID}.zip/" $(DIR)/thin-egress-app.yaml
+	sed -i -e "s/<BUILD_ID>/${BUILD_ID}/g" $(DIR)/thin-egress-app.yaml
+	sed -i -e "s/^Description:.*/Description: \"TEA snapshot, version: ${BUILD_ID} built ${DATE}\"/" $(DIR)/thin-egress-app.yaml
 
-deploy-dependencies: dist/dependencies-$(BUILD_ID).zip
-	aws s3 cp --profile=$(AWS_PROFILE) dist/dependencies-$(BUILD_ID).zip s3://$(CODE_BUCKET)/$(CODE_DIR)/dependencies-$(BUILD_ID).zip
 
-deploy-code: dist/code-$(BUILD_ID).zip
-	aws s3 cp --profile=$(AWS_PROFILE) dist/code-$(BUILD_ID).zip s3://$(CODE_BUCKET)/$(CODE_DIR)/code-$(BUILD_ID).zip
+##############
+# Deployment #
+##############
+# TODO(reweeden): Terraform?
+# TODO(reweeden): Bucket map
 
-deploy-cloudformation: dist/thin-egress-app-$(BUILD_ID).yaml
-	aws cloudformation deploy --region=$(AWS_DEFAULT_REGION) \
-						 --stack-name thin-egress-app-$(BUILD_ID) \
-						 --template-file dist/thin-egress-app-$(BUILD_ID).yaml \
+# Empty targets so we don't re-deploy stuff that is unchanged. Technically they
+# might not be empty, but their purpose is the same.
+# https://www.gnu.org/software/make/manual/html_node/Empty-Targets.html
+
+$(EMPTY)/.deploy-dependencies: $(DIR)/thin-egress-app-dependencies.zip | $(EMPTY)
+	@echo "Deploying dependencies"
+	$(AWS) s3 cp --profile=$(AWS_PROFILE) \
+		$(DIR)/thin-egress-app-dependencies.zip \
+		s3://$(CODE_BUCKET)/$(CODE_PREFIX)dependencies-$(S3_ARTIFACT_TAG).zip
+
+	@echo $(S3_ARTIFACT_TAG) > $(EMPTY)/.deploy-dependencies
+
+$(EMPTY)/.deploy-code: $(DIR)/thin-egress-app-code.zip | $(EMPTY)
+	@echo "Deploying code"
+	$(AWS) s3 cp --profile=$(AWS_PROFILE) \
+		$(DIR)/thin-egress-app-code.zip \
+		s3://$(CODE_BUCKET)/$(CODE_PREFIX)code-$(S3_ARTIFACT_TAG).zip
+
+	@echo $(S3_ARTIFACT_TAG) > $(EMPTY)/.deploy-code
+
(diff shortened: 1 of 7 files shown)


reweeden:
Rename `CONFIG` to something else since there is already a folder named `config`. Better to avoid issues with case insensitive file systems.

Makefile
```diff
@@ -0,0 +1,256 @@
...
+SOURCES := \
+	lambda/app.py \
+	lambda/tea_bumper.py \
+	lambda/update_lambda.py
+
+RESOURCES := \
+	lambda/templates/base.html \
+	lambda/templates/error.html \
+	lambda/templates/profile.html \
+	lambda/templates/root.html
```
reweeden, line 12: Wildcard?

Makefile
```diff
@@ -0,0 +1,256 @@
...
+	cd $(DIR)/terraform && zip ../thin-egress-app-terraform.zip \
+		*.tf \
+		thin-egress-app.yaml \
+		lambda.zip \
+		dependencylayer.zip
+
+##############
+# Deployment #
+##############
+# TODO(reweeden): Terraform?
```
reweeden, line 136: What is terraform used for? Do we need a make target for running deploys with terraform? Probably not...

README.MD
```diff
@@ -195,65 +195,76 @@ builds and code cannot be released without all tests completing successfully.
...
-
-# Make up a filename for code archive:
-DEPENDENCYLAYERFILENAME=tea_depencency_layer.zip
-
-# get the repo
-git clone https://github.com/asfadmin/thin-egress-app
-cd thin-egress-app
+- `zip` for creating zip files
+- `pip` for building the dependency layer
+- `awscli` (optional) for deploying to AWS
```
benbart, line 203: * make (if you want to get _really_ pedantic about it)
* git

Makefile
benbart: The cumulus deploy uses terraform

README.MD
reweeden: Personally, I don't feel the need to be that pedantic :P.

Makefile
reweeden: Yea I found that eventually. I think since the deploy stuff is just for development purposes we don't need to support both terraform and cloudformation. It should be sufficient to just use cloudformation.

README.MD
benbart: I just thought two items looked better than one.

requirements.txt
```diff
@@ -2,7 +2,7 @@ cachetools==5.0.0
 cfnresponse==1.1.2
 chalice==1.26.5
 flatdict==4.0.1
-git+https://github.com/asfadmin/rain-api-core.git@17f5e026749da048553fab825543cbf76c8a4f77
+git+https://github.com/asfadmin/rain-api-core.git@42032ec569a60b4ac9672f4faa81e41b8f6eef27
```
benbart, line 5: I get this: 
`fatal: reference is not a tree: 42032ec569a60b4ac9672f4faa81e41b8f6eef27` 

Choosing other commit hashes works.

reweeden: Yea, I got this error before as well actually. I need to check what this is pointing at since I don't actually remember why I updated the hash in this PR.

Strangely enough I found that upgrading pip made the error go away. Either that or building the dependencies in the docker container, presumably that has a newer version of pip which doesn't encounter this issue. No idea why it would only happen for particular commits...

benbart: That's even more weird. I found I got the same error when I tried to git checkout that hash, independent of the pip install. I didn't try it in another environment though.

reweeden: Actually, this seems to be a dangling commit or something. I can still view it on github but it shows a warning that the commit doesn't belong to the repo. I think I can just undo this change.

https://github.com/asfadmin/rain-api-core/commit/42032ec569a60b4ac9672f4faa81e41b8f6eef27

reweeden: Ahhhh, here's the difference between the pip versions:

- `pip 20.0.2`
```
Cloning https://github.com/asfadmin/rain-api-core.git (to revision 42032ec569a60b4ac9672f4faa81e41b8f6eef27) to /tmp/pip-req-build-k8zsjoat
  Running command git clone -q https://github.com/asfadmin/rain-api-core.git /tmp/pip-req-build-k8zsjoat
  Running command git checkout -q 42032ec569a60b4ac9672f4faa81e41b8f6eef27
  fatal: reference is not a tree: 42032ec569a60b4ac9672f4faa81e41b8f6eef27
```

- `pip 22.0.3`
```
  Cloning https://github.com/asfadmin/rain-api-core.git (to revision 42032ec569a60b4ac9672f4faa81e41b8f6eef27) to /tmp/pip-req-build-gb51fky9
  Running command git clone --filter=blob:none --quiet https://github.com/asfadmin/rain-api-core.git /tmp/pip-req-build-gb51fky9
  Running command git rev-parse -q --verify 'sha^42032ec569a60b4ac9672f4faa81e41b8f6eef27'
  Running command git fetch -q https://github.com/asfadmin/rain-api-core.git 42032ec569a60b4ac9672f4faa81e41b8f6eef27
  Running command git checkout -q 42032ec569a60b4ac9672f4faa81e41b8f6eef27
  Resolved https://github.com/asfadmin/rain-api-core.git to commit 42032ec569a60b4ac9672f4faa81e41b8f6eef27
```

Notice how the new version of pip explicitly fetches the commit:
```
Running command git fetch -q https://github.com/asfadmin/rain-api-core.git 42032ec569a60b4ac9672f4faa81e41b8f6eef27
```

So it can handle detached commits or commits that are otherwise unreachable from a fresh clone of the default branch.

Makefile
```diff
@@ -0,0 +1,256 @@
...
+.PHONY: terraform
+terraform: $(DIR)/thin-egress-app-terraform.zip
+	@echo "Build Terraform zip file for version ${BUILD_ID}"
+
+.PHONY: clean
+clean:
+	rm -rf $(DIR)
+
+$(DIR)/thin-egress-app-dependencies.zip: requirements.txt | $(DIR)/python
+	$(DOCKER_COMMAND) build/dependency_builder.sh "$(DIR)/thin-egress-app-dependencies.zip" "$(DIR)"
```
reweeden, line 83: Might want to delete the old `$(DIR)/python` directory if it exists to make sure we don't zip up left over dependencies from previous builds. i.e. if the change to the requirements.txt file was that a dependency got removed.

build/Jenkinsfile
```diff
@@ -116,55 +122,44 @@ pipeline {
...
-            sh 'sed -i -e "s/<DEPENDENCY_ARCHIVE_PATH_FILENAME>/${CODE_DIR}\\/${DEPENDENCYLAYERFILENAME}/" ./cloudformation/thin-egress-app.yaml'
-            sh 'sed -i -e "s/<BUILD_ID>/${BUILD_ID}/g" ./cloudformation/thin-egress-app.yaml'
-            sh 'sed -i -e "s;^Description:.*;Description: \\"TEA built by Jenkins job ${JOB_NAME}, ${BUILDTAG}\\";" ./cloudformation/thin-egress-app.yaml'
+            sh "echo '>> Building lambda code....'"
+            sh "make code BUILD_ID=${BUILDTAG}"
+            sh "cp ${WORKSPACE}/dist/thin-egress-app-code.zip ${WORKSPACE}/${CODE_ARCHIVE_FILENAME}"
+
+            sh "echo '>> Building CloudFormation yaml....'"
+            sh """ make yaml \
+                      CF_DEFAULT_CODE_BUCKET=${CODE_BUCKET} \
```
reweeden, line 152: Should this actually be `${CODE_BUCKET}${CODE_BUCKET_SUFFIX}`? For prod it doesn't matter but for dev/test deployments the default bucket won't actually be the same bucket where the default code zip and dependency layer are uploaded.

Makefile
```diff
@@ -0,0 +1,246 @@
...
+#####################
+
+# A tag to distinguish between artifacts of the same name. The tag should be
+# different for each build.
+S3_ARTIFACT_TAG = $(DATE_SHORT)
+
+
+# Include custom configuration
+Makefile.config:
+	@echo "It looks like you are building TEA for the first time.\nPlease review the configuration in '$@' and run Make again.\n"
```
benbart, line 40: The newlines aren't unescaping in my console:
```
Makefile:44: Makefile.config: No such file or directory
It looks like you are building TEA for the first time.\nPlease review the configuration in 'Makefile.config' and run Make again.\n
```

Makefile
```diff
@@ -0,0 +1,246 @@
...
+DIR := dist
+EMPTY := $(DIR)/empty
+# Temporary artifacts
+DIST_SOURCES := $(SOURCES:lambda/%=$(DIR)/code/%)
+DIST_RESOURCES := $(RESOURCES:lambda/%=$(DIR)/code/%)
+DIST_TERRAFORM := $(TERRAFORM:terraform/%=$(DIR)/terraform/%)
+
+BUCKET_MAP_OBJECT_KEY := $(CONFIG_PREFIX)bucket-map.yaml
+
+DATE := $(shell date --utc "+%b %d %Y, %T %Z")
```
benbart, line 19: It seems `date` is different on a Mac. When running directly on the mac, this gives me 
```
date: illegal option -- -
usage: date [-jnRu] [-d dst] [-r seconds] [-t west] [-v[+|-]val[ymwdHMS]] ... 
            [-f fmt date | [[[mm]dd]HH]MM[[cc]yy][.ss]] [+format]
```

I wonder if maintaining the ability to do both is not worth it?

build/dependency_builder.sh
```diff
@@ -1,50 +1,30 @@
...
-
-echo "Updating Pip..."
-python3.8 -m pip install -U pip
-echo "Installing setuptools..."
-python3.8 -m pip install --upgrade setuptools
 echo "Installing ${WORKSPACE}/requirements.txt"
-python3.8 -m pip install -r "${WORKSPACE}"/requirements.txt --target .
+python3.8 -m pip install \
+  --upgrade \
+  -r "${WORKSPACE}/requirements.txt" \
```
benbart, line 13: When building in an amazonlinux:2 container:

```
Collecting flatdict==4.0.1 (from -r /tea/requirements.txt (line 4))
  Downloading https://files.pythonhosted.org/packages/3e/0d/424de6e5612f1399ff69bf86500d6a62ff0a4843979701ae97f120c7f1fe/flatdict-4.0.1.tar.gz
    Complete output from command python setup.py egg_info:
    setuptools 39.2 or greater required for installation
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-b8uibfdj/flatdict/
```

I'm going to try in the lambci/lambda:build-python3.8  or some other py2.8 container, but pulling is slow right now.

Makefile
reweeden: Ok. interesting

Makefile
reweeden: Well I could do `-u` instead of `--utc`. Or drop that arg all together. The timezone isn't really important in the grand scheme of things.

Makefile
reweeden: Also, is it printing the 'no such file or directory' error before the messages? For me it's printing after:

```
It looks like you are building TEA for the first time.
Please review the configuration in 'Makefile.config' and run Make again.

Makefile:46: Makefile.config: No such file or directory
make: *** [Makefile:44: Makefile.config] Error 1
```

I haven't found a way to get make to just stop there without raising an error. Either I make it error out or it continues trying to execute the `build` target right after.

Makefile
reweeden: Let me know if that worked and I'll resolve this comment

Makefile
```diff
@@ -0,0 +1,248 @@
...
+	@echo "It looks like you are building TEA for the first time."
+	@echo "Please review the configuration in '$@' and run Make again."
+	@echo
+	@cp --no-clobber Makefile.config.example $@
+	@exit 1
+
+include Makefile.config
+
+
+.DEFAULT_GOAL := build
```
reweeden, line 49: Should add an `all` target that is an 'alias' for `build` as that's a pretty standard name for the default target.

Makefile
```diff
@@ -0,0 +1,250 @@
...
+.SECONDARY: $(DIST_RESOURCES)
+$(DIST_RESOURCES): $(DIR)/code/%: lambda/%
+	@mkdir -p $(@D)
+	cp $< $@
+
+.SECONDARY: $(DIST_SOURCES)
+$(DIST_SOURCES): $(DIR)/code/%: lambda/%
+	@mkdir -p $(@D)
+	cp $< $@
+	sed -i "s;<BUILD_ID>;${BUILD_ID};g" $@
```
benbart, line 100: On Mac, I'm getting:
```
$ make
cp lambda/app.py dist/code/app.py
sed -i "s;<BUILD_ID>;095f7f7;g" dist/code/app.py
sed: 1: "dist/code/app.py": extra characters at the end of d command
make: *** [dist/code/app.py] Error 1
```

Doing `sed -i '' "s;<BUILD_ID>;095f7f7;g" dist/code/app.py` works on the Mac, but breaks linux.

reweeden: What if you remove the space after `-i` like so: `sed -i"" ...`. According to this post, that will work on both linux and OS X 10.9+.

https://stackoverflow.com/a/4247319

Probably the most foolproof way of fixing this would be to do the substitutions in a python script instead.

benbart: Huh. Removing the space allows it to work in linux. But breaks it in MacOS.

reweeden: Python script it is then.

build/dependency_builder.sh
benbart: This works in a `lambci/lambda:build-python3.8` container but I still get the above error in `amazonlinux:2` with python3.8 installed.

Makefile
benbart: I haven't run into this problem in recent runnings of the makefile

reweeden: Ok cool. I'll consider this to be resolved for now then

Makefile
```diff
@@ -0,0 +1,250 @@
...
+# Temporary artifacts
+DIST_SOURCES := $(SOURCES:lambda/%=$(DIR)/code/%)
+DIST_RESOURCES := $(RESOURCES:lambda/%=$(DIR)/code/%)
+DIST_TERRAFORM := $(TERRAFORM:terraform/%=$(DIR)/terraform/%)
+
+BUCKET_MAP_OBJECT_KEY := $(CONFIG_PREFIX)bucket-map.yaml
+
+DATE := $(shell date -u "+%b %d %Y, %T %Z")
+DATE_SHORT := $(shell date -u "+%Y%m%dT%H%M%S")
+BUILD_ID := $(shell git rev-parse --short HEAD 2>/dev/null || echo "SNAPSHOT")
```
reweeden, line 21: Might be a good idea to have a `version.txt` file checked into the repo so that we can know an approximate version even outside of a git environment. For example if someone were to download the code zip from the releases page.

Makefile
```diff
@@ -0,0 +1,251 @@
...
+	@mkdir -p $(@D)
+	cp $< $@
+	python3 scripts/sed.py -i $@ "<BUILD_ID>" "${BUILD_ID}"
+
+$(DIR)/thin-egress-app-code.zip: $(DIST_SOURCES) $(DIST_RESOURCES)
+	@mkdir -p $(DIR)/code
+	cd $(DIR)/code && zip -r ../thin-egress-app-code.zip .
+
+$(DIR)/bucket-map.yaml: config/bucket-map-template.yaml
+	cp $< $@
```
reweeden, line 107: Maybe use `cp -i` here to prompt for overwrites?

Makefile.config.example
```diff
@@ -0,0 +1,69 @@
...
+CONFIG_PREFIX := $(STACK_NAME)/
+# Uncomment the following line to use an existing bucket map rather than uploading one
+# BUCKET_MAP_OBJECT_KEY :=
+
+# String that will be prepended to bucket names after looking them up in the bucket map
+BUCKETNAME_PREFIX :=
+
+# These secrets are required but not managed by the Makefile.
+# For full deployments use `build/deploy.sh`
+JWT_KEY_SECRET_NAME := REQUIRED!
```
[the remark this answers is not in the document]
benbart: Good idea. Add to the issues if you don't want to deal with it immediately.

Makefile
benbart: Yes, on the Mac, the `Makefile:44...` prints out before the message. On my two linux containers, it's as you pasted above.

Makefile
```diff
@@ -0,0 +1,251 @@
...
+# different for each build.
+S3_ARTIFACT_TAG = $(DATE_SHORT)
+
+
+# Include custom configuration
+Makefile.config:
+	@echo "It looks like you are building TEA for the first time."
+	@echo "Please review the configuration in '$@' and run Make again."
+	@echo
+	@cp --no-clobber Makefile.config.example $@
```
benbart, line 43: On Mac, I'm getting
```
cp: illegal option -- -
usage: cp [-R [-H | -L | -P]] [-fi | -n] [-apvXc] source_file target_file
       cp [-R [-H | -L | -P]] [-fi | -n] [-apvXc] source_file ... target_directory
make: *** [Makefile.config] Error 64
```
Simply `cp -n ...` works on Mac. But reportedly doesn't work on RedHat.

Makefile
reweeden: I'll add a blank line above as well then

Makefile
reweeden: :sweat: Mac doesn't like human readable argument names apparently. Well I found a little trick that avoids the argument completely https://superuser.com/a/847892. In the worst case the argument can just be removed here since Make *shouldn't* be invoking the target if the file already exists.

Makefile.config.example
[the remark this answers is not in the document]
reweeden: Added

Makefile
```diff
@@ -0,0 +1,246 @@
...
+	@mkdir -p $(@D)
+	cp $< $@
+	sed -i "s/<BUILD_ID>/${BUILD_ID}/g" $@
+
+$(DIR)/thin-egress-app-code.zip: $(DIST_SOURCES) $(DIST_RESOURCES)
+	@mkdir -p $(DIR)/code
+	cd $(DIR)/code && zip -r ../thin-egress-app-code.zip .
+
+$(DIR)/bucket-map.yaml: config/bucket-map-template.yaml
+	cp $< $@
```
benbart: On Mac, I get:
```
$ make deploy
cp -i  dist/bucket-map.yaml
usage: cp [-R [-H | -L | -P]] [-fi | -n] [-apvXc] source_file target_file
       cp [-R [-H | -L | -P]] [-fi | -n] [-apvXc] source_file ... target_directory
make: *** [dist/bucket-map.yaml] Error 64
```

reweeden: Hmm, so Mac wants `-fi` and not just `-i` on its own?

reweeden: Oh wait, no for some reason the `$<` prerequisite variable was blank??

reweeden: Oh, it's because I changed it to a order-only dependency.

Makefile
```diff
@@ -0,0 +1,252 @@
...
+	@mkdir -p $(@D)
+	cp $< $@
+	python3 scripts/sed.py -i $@ "<BUILD_ID>" "${BUILD_ID}"
+
+$(DIR)/thin-egress-app-code.zip: $(DIST_SOURCES) $(DIST_RESOURCES)
+	@mkdir -p $(DIR)/code
+	cd $(DIR)/code && zip -r ../thin-egress-app-code.zip .
+
+$(DIR)/bucket-map.yaml: | config/bucket-map-template.yaml
+	cp -i $| $@
```
reweeden, line 108: This probably makes mores sense TBH
```suggestion
$(DIR)/bucket-map.yaml:
	cp -i config/bucket-map-template.yaml $@
```

build/Jenkinsfile
reweeden: Probably just leave it for now. The stuff uploaded to the suffixed bucket is supposed to be for testing only, the artifacts for dev/test builds aren't ever uploaded for distribution. Besides, I'm rewriting the Jenkinsfile with GitHub actions anyways.

mckadesorensen (approved):

PR #459, opened 2022-02-14
PR-3080 Add github actions

reweeden:
This changes how the release process will work a little bit since the pipeline will be triggered by the act of creating a release on GitHub, rather than the current way of triggering a build in Jenkins and then making a release from the tag it creates.


.github/workflows/release.yml
```diff
@@ -0,0 +1,114 @@
...
+          SKIP_TESTS: ${{ secrets.SKIP_TESTS }}
+        if: "${{ env.SKIP_TESTS == '' }}"
+        run: echo "::set-output name=defined::true"
+
+  test-e2e:
+    needs:
+      - build
+      - check-tests
+    if: needs.check-tests.outputs.run-tests == 'true'
+    uses: ./.github/workflows/re-test-e2e.yml
```
reweeden, line 38: Maybe it makes sense not to run the tests again on release, since they would have run already on push to `devel` and `master`

.github/workflows/re-test-e2e.yml
```diff
@@ -0,0 +1,217 @@
...
+      EDL_APP_PASSWORD:
+        required: false
+
+jobs:
+  # Deploy to the test environment and run end to end tests
+  test-end-to-end:
+    runs-on: ubuntu-latest
+    environment: ${{ inputs.environment }}
+    env:
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
```
Jlrine2, line 68: I don't know if it is still an issue but I have had issues with using the CLI with environment variable authentication config. there is an action that setsup credentials for you
```
      - uses: aws-actions/configure-aws-credentials@v1
        with:
          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
          aws-region: ${{ secrets.AWS_REGION }}
          role-to-assume: ${{ secrets.AWS_ROLE_ARN }}
```

.github/workflows/re-test-e2e.yml
```diff
@@ -0,0 +1,217 @@
...
+             echo ">> Updating URS secret '$URS_AUTH_CREDS_SECRET_NAME'"
+
+             aws secretsmanager put-secret-value \
+                --secret-id "$URS_AUTH_CREDS_SECRET_NAME" \
+                --secret-string file://secret.txt
+          fi
+
+          shred secret.txt
+
+      - name: Update CloudFormation stack description
```
Jlrine2, line 164: what is the intent here? if it is to identify this build would it be better accomplished by taggin the stack in  the deploy command?

reweeden: Good point. This is just how it was being done in Jenkins and I just blindly ported it over. I like the idea of just doing it in the deploy command though, I think that saves a lot of hassle.

.github/workflows/re-test-e2e.yml
reweeden: Seems to have been working fine for me so far  :shrug:

.github/workflows/re-test-e2e.yml
reweeden: Ehhhh, it doesn't seem that this is possible? https://docs.aws.amazon.com/cli/latest/reference/cloudformation/deploy/index.html

You can only set stack name and override parameters, but description is not a parameter.

.github/workflows/release.yml
```diff
@@ -0,0 +1,157 @@
+# This workflow runs whenever a GitHub release is created. It gets the version
+# number from the git tag, builds and tests TEA, and then uploads the artifacts
+# to the public bucket for distribution.
+name: Release
+
+on:
+  release:
+    types:
+      - released
```
reweeden, line 9: This event type doesn't trigger on prereleases. There are some event types that do trigger on prereleases which I could use instead if we want that.

.github/workflows/re-build.yml
```diff
@@ -0,0 +1,102 @@
...
+          TIMESTAMP=$(TZ=America/Anchorage date)
+
+          cat > Makefile.config << EOF
+          BUILD_ID := ${{ needs.variables.outputs.version }}
+
+          CF_DEFAULT_CODE_BUCKET := $CODE_BUCKET
+          CF_DEFAULT_DEPENDENCY_ARCHIVE_KEY := ${{ env.CODE_PREFIX }}${{ needs.variables.outputs.code-zip }}
+          CF_DEFAULT_CODE_ARCHIVE_KEY = ${{ env.CODE_PREFIX }}${{ needs.variables.outputs.dependency-zip }}
+          CF_BUILD_VERSION := $(BUILD_ID)
+          CF_DESCRIPTION := TEA version ${{ needs.variables.outputs.version}} (${GITHUB_SHA:0:7}) built by GitHub Actions on $TIMESTAMP.
```
reweeden, line 74: Build version seems to be empty and description is not being replaced.

.github/workflows/README.md
```diff
@@ -0,0 +1,52 @@
+You can stop the integration tests from running by setting `SKIP_TESTS` in your
+repository secrets to `true`. All other configuration is done through GitHub
+environments.
```
reweeden, line 3: Probably better to make it default to forks not trying to run the e2e tests as they are very specific to ASF's setup

.github/workflows/re-build.yml
```diff
@@ -0,0 +1,96 @@
...
+      version: ${{ steps.step1.outputs.version }}
+      code-zip: ${{ steps.step1.outputs.code-zip }}
+      dependency-zip: ${{ steps.step1.outputs.dependency-zip }}
+      cloudformation-yaml: ${{ steps.step1.outputs.cloudformation-yaml }}
+      terraform-zip: ${{ steps.step1.outputs.terraform-zip }}
+
+    steps:
+      - id: step1
+        run: |
+          VERSION=${GITHUB_REF_NAME#*.}
```
reweeden, line 38: This takes the tag name and strips off anything before the first `.`. We may want to change this depending on how we plan to tag our releases going forward. I chose this so it would work with the tags created by jenkins which are of the format: `tea*-build.xxx`.

.github/workflows/README.md
```diff
@@ -0,0 +1,85 @@
...
+    the following secrets are also needed:
+    - `URS_CLIENT_ID`
+    - `EDL_APP_UID`
+    - `EDL_APP_PASSWORD`
+
+### Setting up AWS Credentials
+- `AWS_ACCESS_KEY_ID`
+- `AWS_SECRET_ACCESS_KEY`
+- (optional) `AWS_ROLE_ARN`
+- (optional) `AWS_REGION`
```
benbart, line 42: `Upload success labels` fails when AWS_REGION is not set

.github/workflows/README.md
```diff
@@ -0,0 +1,85 @@
+# GitHub Actions Workflows
```
benbart, line 1: Not sure its within the scope of this document, but it should be included somewhere:

Codecov is required for build to happen and needs to be set up for the repo.

Maybe a "forking" section needs to be added to the main README, with a link to this document?

reweeden: Hmm. Good point. Does it cause an error? I think as long as a repo is public it can just publish results to codecov. You don't need to setup a token or anything (that's just for private repos).

benbart: It causes this error: https://github.com/benbart/thin-egress-app/actions/runs/1878327064

```
[2022-02-21T21:08:01.224Z] ['error'] There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 404 - Not Found - {'detail': ErrorDetail(string='Could not find a repository, try using repo upload token', code='not_found')}
Error: Codecov: Failed to properly upload: The process '/home/runner/work/_actions/codecov/codecov-action/v2/dist/codecov' failed with exit code 255
```
I had to let codecov know about my repo before it would accept uploads from it.

.github/workflows/re-build.yml
benbart: Thanks for this explanation

.github/workflows/README.md
reweeden: Ok, makes sense

.github/workflows/README.md
reweeden: So it only failed in the last step when trying to upload the labels, and not in the publish step when uploading the actual artifacts?

.github/workflows/README.md
reweeden: I added a note to the readme and added a link to it from the main readme

PR #464, opened 2022-02-23
Improve support for building local dependencies

reweeden:
Some changes I needed to do to be able to build TEA with a local development version of rain api core.

### DIFF
--- a/Makefile.config.example
+++ b/Makefile.config.example
@@ -67,8 +67,14 @@ AWS_PROFILE := default
 ########
 # Misc #
 ########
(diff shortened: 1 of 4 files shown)


mckadesorensen (approved):

Makefile.config.example
```diff
@@ -67,8 +70,14 @@ AWS_PROFILE := default
 ########
 # Misc #
 ########
+# Add any local dependencies here that are required for the dependency layer to be built.
```
benbart, line 73: It would be ever so slightly more helpful if these two options were in the same order as they are in the readme.

PR #469, opened 2022-03-01
Merge devel into master

reweeden:
Make sure that `devel` isn't deleted automatically when the pr is merged!


reweeden:
I was afraid that was going to happen

PR #491, opened 2022-03-22
General cleanup

reweeden:
A bunch of mostly code style things I wanted to fix (see commit messages for a list)

### DIFF
--- a/lambda/app.py
+++ b/lambda/app.py
@@ -595,9 +595,9 @@ def version():
     log.info("Got a version request!")
     version_return = {'version_id': '<BUILD_ID>'}
 
-    # If we've flushed, lets return the flush time.
-    if os.getenv('BUMP'):
-        version_return['last_flush'] = os.getenv('BUMP')
+    flush_time = os.getenv('BUMP')
+    if flush_time:
+        version_return['last_flush'] = flush_time
 
     return json.dumps(version_return)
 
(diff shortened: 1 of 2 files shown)


benbart:
I don't like the wide-ranging cosmetic changes like changing all the quotes to single quotes. Mainly because it obscures the history/blame. Also because it doesn't matter.

reweeden:
That's why they're all in their own commit, so that it's obvious what changed. You can just look at the commit message and see that the commit didn't change anything of substance and so just skip it and go to the next one. It's only a problem when you mix these changes with actual code changes and then have a single commit that does both.

benbart:
It's still a problem because tools that show you the most recent commit, show you the most recent commit. So if you want to see what's up, you need to go digging.

reweeden:
I don't see how this is different from any other refactor where a piece of code is moved from one place to another. At the end of the day, if you really want to know who the first person was to write a particular line of code you will always have to go digging. But IMO, that's not even important information except in maybe a tiny fraction of cases.

I guess I don't understand what your use case is for viewing the blames, since I don't seem to have ever encountered any annoyances with this.

mckadesorensen:
I agree with Ben on this one. I don't think the quote changes are worth messing up the git history. I find it useful to be able to look at the commit messages in my IDE. If there were more changes, I would agree with it a bit more.

reweeden:
It doesn't mess up the history. Besides, that is only 1 commit out of 5 in this PR. I can drop it if people really feel that strongly about it, but IMO having a consistently formatted code base is much more important than "not making a commit because it will appear in the history". The code is going to continue to change, and whenever that happens there will be more commits. That's just how git works.

PR #493, opened 2022-03-22
PR-3130 Add full dependency locking

reweeden:
Locks the full dependency tree with pip-tools so we only need to specify top level dependencies, but still get deterministic builds.

Some things we could tweak:
- The comments showing where dependencies are included from can be disabled with `--no-annotate`
- Packages like setuptools and pip which are currently not pinned can be pinned with `--allow-unsafe`. This will apparently become default in future pip-tools versions
- I decided not to bother with pip tools for the mkdocs requirements and end to end test requirements, although we could do it for those as well. It just seemed like an aweful lot of requirements.* files.
- ~~We could add the pip-compile command to the Makefile~~ **done**
- ~~We could run the pip-compile command inside of the lambda-ci docker container. It might be the case that on MacOS, the compiled package list is different than on linux, so this would ensure that we always get exactly the right list for AWS lambda.~~ **done**
- We might need to change pyup bot settings to point to the new requirements files. I'm not even sure if pyup knows how to handle pip-tools but I know that dependabot does.


README.MD
```diff
@@ -39,12 +39,38 @@ releases see the [GitHub Actions README](.github/workflows/README.md).*
...
+
+```bash
+# Create virtual environment for development
+python3 -m venv .venv
+# Activate virtual environment
+source .venv/bin/activate
+
+# Install requirements to the virtual environment.
+# Make sure you see '(.venv)' in your shell prompt before running this!
+pip install -r requirements/requirements.txt -r requirements/requirements-dev.txt
```
benbart, line 69: When I naively try this on a Mac (Python 3.8.9), it fails to install cryptography for lack of Rust compiler.

benbart: It works in a docker image.

reweeden: Hmm, true. The lock files are environment specific, so for mac you'd have to recompile the .in file to a `requirements-mac.txt` file

reweeden: I went ahead and updated the instructions to mention the environment dependence, and I changed the make command to run the locking inside of a docker container to make sure the pinned versions are always generated for the right environment.

.github/workflows/re-test-e2e.yml
```diff
@@ -186,7 +186,7 @@ jobs:
           URS_USERNAME: ${{ secrets.URS_USERNAME }}
           URS_PASSWORD: ${{ secrets.URS_PASSWORD }}
         run: |
-          pip install -r requirements-dev.txt GDAL==3.0.4
+          pip install -r requirements/requirements-test.txt GDAL==3.0.4
```
Jlrine2, line 189: can GDAL not be put in the requirements files?

Jlrine2 (approved):

.github/workflows/re-test-e2e.yml
reweeden: It could. It's a bit weird because of a few things:
1. I want it to be optional since it's only used for 1 or 2 tests and GDAL is a big heavy dependency that is annoying to install, so I don't want to force that on everyong wanting to develop TEA
2. The version of the GDAL python package you install needs to match the version of the c-library installed on the system. I don't currently have a solution to this, it's possible that this may just randomly break if the version of GDAL distributed through apt is updated.

README.MD
```diff
@@ -39,12 +39,52 @@ releases see the [GitHub Actions README](.github/workflows/README.md).*
...
+[pip-tools](https://pypi.org/project/pip-tools/). For housekeeping purposes
+these are all located in the `requirements` directory. If you need to add a
+runtime dependency, put it into `requirements/requirements.in` and then run
+`make lock` to generate the `requirements/requirements.txt` file. This will
+run `pip-compile` in a docker container to ensure the file is compiled for a
+lambda compatible environment.
+
+```bash
+# Update the pinned requirements files
+make lock
```
benbart, line 62: Is the word `lock` a common convention for this? Seems like it would be somewhat ambiguous when separated from this documentation if not.

README.MD
```diff
@@ -39,12 +39,52 @@ releases see the [GitHub Actions README](.github/workflows/README.md).*
...
+```
+
+Note that the pinned `requirements.txt` files are environment specific. i.e.
```
benbart, line 65: Do I understand correctly that the `make lock` above is to make requirements.txt files for commit to github and this section is for creating requirements files for local dev?

benbart (approved):

PR #526, opened 2022-04-11
Refactor tea middleware

reweeden:
Chalice has support for middleware so we don't need to do the hack with subclassing

### DIFF
--- a/lambda/app.py
+++ b/lambda/app.py
@@ -10,6 +10,7 @@ from urllib.error import HTTPError
 from urllib.parse import quote_plus, urlencode, urlparse
 
 import cachetools
+import chalice
 import flatdict
 from botocore.config import Config as bc_Config
 from botocore.exceptions import ClientError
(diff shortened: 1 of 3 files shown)


lambda/app.py
```diff
@@ -101,36 +103,44 @@ def get_black_list():
...
-        origin_request_id = event.get('headers', {}).get('x-origin-request-id')
-        log_context(route=resource_path, request_id=context.aws_request_id, origin_request_id=origin_request_id)
-        # JWT_MANAGER.get_profile_from_headers() below generates log messages, so the above log_context() sets the
-        # vars for it to use while it's doing the username lookup
-        user_profile = JWT_MANAGER.get_profile_from_headers(event.get('headers'))
-        if user_profile is not None:
-            log_context(user_id=user_profile.user_id)
+@app.middleware('http')
+def set_log_context(event: chalice.app.Request, get_response):
+    JWT_MANAGER.black_list = get_black_list()
```
Jlrine2, line 111: this line feels a bit out of place, the black_list and log context are very unrelated, not sure if want a more generic function name, or more functions.

Jlrine2 (approved):

mckadesorensen (approved):

lambda/app.py
reweeden: Kinda true, but it needs to be there because the function calls `get_profile_from_headers` which needs the blacklist

Jlrine2 (approved):

PR #614, opened 2022-05-31
Updated ACLs in release script

mckadesorensen:


.github/workflows/release.yml
```diff
@@ -71,18 +71,18 @@ jobs:
         env:
           S3_PATH_PREFIX: s3://${{ env.CODE_BUCKET }}/${{ env.CODE_PREFIX }}
         run: |
-          aws s3 cp ./dependency-layer/thin-egress-app-dependencies.zip ${S3_PATH_PREFIX}${{ needs.build.outputs.dependency-zip }}
-          aws s3 cp ./code/thin-egress-app-code.zip ${S3_PATH_PREFIX}${{ needs.build.outputs.code-zip }}
-          aws s3 cp ./cloudformation/thin-egress-app.yaml ${S3_PATH_PREFIX}${{ needs.build.outputs.cloudformation-yaml }}
-          aws s3 cp ./terraform/thin-egress-app-terraform.zip ${S3_PATH_PREFIX}${{ needs.build.outputs.terraform-zip }}
+          aws s3 cp ./dependency-layer/thin-egress-app-dependencies.zip ${S3_PATH_PREFIX}${{ needs.build.outputs.dependency-zip }}  --acl bucket-owner-full-control --acl public-read
```
reweeden, line 74: I'm not sure `bucket-owner-full-control` is needed. It's not present in the Jenkinsfile

mckadesorensen: If we want to be able to edit the files at all in the future it is needed. If it isn't added then the `11` account has ownership and not the `7` account.

benbart (approved):

.github/workflows/release.yml
benbart: It should have been in the Jenkinsfile, it turns out.

reweeden: That might depend on how the credentials are set up? I guess it doesn't hurt to add it

benbart: The creds used by actions are apparently for the `11` account, yet the bucket is in the `7` account. The bucket has a policy attached to allow the `11` account to upload to it. Is there a reason to use creds for `11` instead of `7` here?

PR #642, opened 2022-10-14
PR-3244 Add s3 credentials endpoint

reweeden:
Adds and endpoint to dispense temporary s3 credentials like the cumulus-distribution module does. One key difference is that I'm calling STS directly from TEA whereas cumulus-distribution goes through the NGAP provided `gsfc-ngap-sh-s3-sts-get-keys` lambda function. I'm not sure why cumulus-distribution uses that lambda, but if we can get away with not calling it then I think that's much preferred as it make the code simpler and probably faster too. Either way we have to support calling sts directly so that TEA can still function outside of NGAP.

TODO:
- [x] Update documentation

### DIFF
--- a/cloudformation/thin-egress-app.yaml
+++ b/cloudformation/thin-egress-app.yaml
@@ -696,6 +696,15 @@ Resources:
       PathPart: '{proxy+}'
       RestApiId: !Ref EgressApiGateway
 
+  EgressApiResourceS3Credentials:
+    Type: AWS::ApiGateway::Resource
+    DependsOn:
+      - EgressApiGateway
+    Properties:
+      ParentId: !GetAtt EgressApiGateway.RootResourceId
+      PathPart: 's3credentials'
+      RestApiId: !Ref EgressApiGateway
+
   EgressApiResourceProfile:
     Type: AWS::ApiGateway::Resource
     DependsOn:
@@ -881,6 +890,27 @@ Resources:
         'method.request.header.Cookie': true
         'method.request.header.X-urs-access-token': true
       ResourceId: !Ref EgressApiResourceDynamicUrl
+      RestApiId: !Ref EgressApiGateway
+
+  EgressAPIMethodS3Credentials:
+    Type: AWS::ApiGateway::Method
+    Properties:
+      ApiKeyRequired: false
+      AuthorizationType: 'NONE'
+      HttpMethod: 'GET'
+      Integration:
+        IntegrationHttpMethod: 'POST'
+        IntegrationResponses:
+          - StatusCode: 200
+          - StatusCode: 401
+        Type: 'AWS_PROXY'
+        Uri: !Sub "arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${EgressLambda.Arn}/invocations"
+      OperationName: 's3credentials view'
+      RequestParameters:
+        'method.request.header.Cookie': true
+        'method.request.header.X-urs-access-token': true
+        'method.request.header.X-urs-app-name': false
+      ResourceId: !Ref EgressApiResourceS3Credentials
       RestApiId: !Ref EgressApiGateway
 
   EgressAPIMethodProfile:
--- a/cloudformation/thin-egress-app.yaml
+++ b/cloudformation/thin-egress-app.yaml
@@ -16,6 +16,7 @@ Metadata:
           - HtmlTemplateDir
           - StageName
           - EnableApiGatewayLogToCloudWatch
+          - EnableS3CredentialsEndpoint
       -
         Label:
           default: Domain Settings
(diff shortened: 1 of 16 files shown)


lambda/app.py
```diff
@@ -910,6 +912,71 @@ def dynamic_url():
     return try_download_from_bucket(entry.bucket, entry.object_key, user_profile, custom_headers)
 
 
+@app.route('/s3credentials', methods=['GET'])
+@with_trace(context={})
+def s3credentials():
```
reweeden, line 917: Should this be optional? We could either return an error response if the feature is not enabled, or not even deploy the endpoint in the cloudformation or both.

lambda/app.py
```diff
@@ -910,6 +912,71 @@ def dynamic_url():
...
+            "requestid": get_request_id(),
+        }
+        headers = {}
+        return make_html_response(template_vars, headers, 401, "error.html")
+    log_context(user_id=user_profile.user_id)
+
+    policy = b_map.to_iam_policy(user_profile.groups)
+    log.debug("policy: %s", policy)
+
+    app_name = app.current_request.headers.get("app-name", "")
```
reweeden, line 941: Different name for the app name header?

lambda/app.py
```diff
@@ -910,6 +912,71 @@ def dynamic_url():
...
+        body=json.dumps(creds, default=str),
+        status_code=200,
+    )
+
+
+@with_trace()
+def get_role_session_name(user_id: str, app_name: str):
+    # https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html#API_AssumeRole_RequestParameters
+    if not re.match(r"[\w+,.=@-]*", app_name):
+        app_name = ""
```
reweeden, line 961: Could just strip out invalid characters.

lambda/app.py
mattp0: I think if it is going to be optional the endpoint should not be deployed.. But NASA does want temp S3 access. Is this specifically to be used with non cumulus datasets that users want direct s3 access to?

lambda/app.py
mattp0: Would this be the `Host` header?

reweeden: I think `Host` just contains the hostname that the client thinks it's connecting to so the server can return the correct website in the case of multiple sites being served by the same machine. The app name here is just an optional additional identifier that can be added to the role_session_name so that it shows up in the s3 access logs. In cumulus-distribution the header is called `EDL-Client-Name` which we could copy for consistency, but IMO it's a bad name and has nothing to do with EDL at all.

tests/test_app.py
```diff
@@ -1296,6 +1296,60 @@ def test_dynamic_url_bearer_auth(
     assert response.headers == {}
 
 
+@mock.patch(f"{MODULE}.get_s3_credentials", autospec=True)
```
mattp0, line 1299: Idk why but more than 1 decorator seems crazy..

what about the `mocker` 
```python
def test_s3credentials(
    mock_get_profile,
    mock_get_yaml_file,
    mock_get_s3_credentials,
    mocker,
    resources,
    user_profile,
    client
):
    mocker.patch(f"{MODULE}.get_s3_credentials", autospec=True)
    mocker.patch(f"{MODULE}.get_yaml_file", autospec=True)
    mocker.patch(f"{MODULE}.JwtManager.get_profile_from_headers", autospec=True)
    mocker.patch(f"{MODULE}.b_map", None)
```

Just recalling when we were working on Palsar

lambda/app.py
reweeden: I'm imagining this parameter being used if there's an app making requests on users behalf, such as the burst extractor. Then it could set this header to some consistent value and you'd be able to tell in the s3 access logs not only which user downloaded the data, but how they downloaded it and if it was actually done indirectly by a proxy app.

tests/test_app.py
reweeden: Yea but I'd have to refactor the entire test suite

lambda/app.py
reweeden: It should work for cumulus and NGAP as well.

lambda/app.py
```diff
@@ -910,6 +912,71 @@ def dynamic_url():
...
+
+    return f"{user_id}@{app_name}"[:64]
+
+
+@with_trace()
+def get_s3_credentials(user_id: str, role_session_name: str, policy: dict):
```
reweeden, line 967: Should we be doing any caching?

lambda/app.py
```diff
@@ -910,6 +912,71 @@ def dynamic_url():
...
+
+    user_profile = JWT_MANAGER.get_profile_from_headers(app.current_request.headers)
+    if user_profile is None:
+        template_vars = {
+            "contentstring": "You must log in first",
+            "title": "Unauthorized",
+            "requestid": get_request_id(),
+        }
+        headers = {}
+        return make_html_response(template_vars, headers, 401, "error.html")
```
reweeden, line 935: We'll want to add the whole redirect thing here if the user isn't logged in. Question is, do we need to support bearer tokens?

mattp0 (dismissed):

cloudformation/thin-egress-app.yaml
```diff
@@ -300,6 +309,7 @@ Parameters:
 Conditions:
   DomainNameIsSet: !Not [ !Equals [ !Ref DomainName, "" ] ]
   ApiGatewayLogToCloudWatchIsSet: !Equals [ !Ref EnableApiGatewayLogToCloudWatch, "True" ]
+  S3CredentialsEndpointIsSet: !Equals [ !Ref EnableS3CredentialsEndpoint, "True" ]
```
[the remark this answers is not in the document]
reweeden: Yea default is False, but this line is not where the default is defined. This is just converting the string parameter to a boolean condition.

lambda/app.py
reweeden: Refactored the auth logic so it can just be plugged in here. Therefore, bearer tokens will also be supported.

mattp0 (approved):

PR #644, opened 2022-10-21
PR-3244 S3 credential fix

reweeden:
The main fix comes from updating the rain-api-core dependency to the version that will dedupe statement resources

### DIFF
--- a/lambda/app.py
+++ b/lambda/app.py
@@ -1011,7 +1011,10 @@ def get_s3_credentials(user_id: str, role_session_name: str, policy: dict):
         RoleSessionName=role_session_name,
         ExternalId=user_id,
         DurationSeconds=3600,
-        Policy=json.dumps(policy)
+        # NOTE: Policy max size is 2048 characters which is quite limiting.
+        # TODO(reweeden): We'll need to figure out how to accomodate large
+        # bucket maps that will push us over this limit.
+        Policy=json.dumps(policy, separators=(",", ":"))
     )
     return response["Credentials"]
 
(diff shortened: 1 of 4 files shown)


lambda/app.py
```diff
@@ -976,6 +976,19 @@ def s3credentials():
     policy = b_map.to_iam_policy(groups)
     log.debug("policy: %s", policy)
 
+    if policy is None:
+        template_vars = {
+            "contentstring": "You do not have permission to access any data.",
+            "title": "Could not access data",
```
mckadesorensen, line 982: I don't think this really matters but it seems weird to have `contentstring` before `title`

tests/test_app.py
```diff
@@ -1441,6 +1441,34 @@ def test_s3credentials_unauthenticated(
...
+):
+    mock_get_yaml_file.return_value = {}
+    mock_get_profile.return_value = user_profile
+
+    client.http.get("/s3credentials")
+
+    mock_make_html_response.assert_called_once_with(
+        {
+            "contentstring": "You do not have permission to access any data.",
+            "title": "Could not access data",
```
mckadesorensen, line 1463: Again, a little weird

mckadesorensen (approved):

lambda/app.py
reweeden: Yea, unfortunately this is the order it's in everywhere. I'd love to refactor this whole file and fix all sorts of stuff like this, but last time I tried to do that I got blocked because people didn't like that refactoring would cause additional commits to be created in the history...

mattp0 (approved):

PR #645, opened 2022-10-21
PR-3244 Use jinja2 for CloudFormation variables

reweeden:
This lets us do the random deployment ID in a more elegant way and opens the door to doing other rendering tasks in python in the future. In particular I'm thinking of rendering a markdown README file for the s3credentials endpoint to an HTML template.

TODO:
- [x] Make sure it works on MacOS


mattp0 (dismissed):
Works on Mac

mattp0 (approved):
Can we remove the sed script as well with this PR? Can always get it back later

reweeden:
It's still used to inject the build ID into `app.py` for the `/version` endpoint. This could probably be refactored as well, but I didn't do it in this PR.

PR #650, opened 2022-10-25
PR-3244 Add s3credentialsREADME endpoint

reweeden:
I decided to reuse the s3access.md markdown from our tea-docs page for the `/s3credentialsREADME` so that we don't need to maintain two separate docs. This means the one on our tea-docs page will always be up to date for the latest release and the one at `/s3credentialsREADME` will be correct for the version of TEA that's deployed in case we change how the endpoint works at some point.


docs/vision.md
```diff
@@ -1,3 +1,4 @@
+# Vision
```
mattp0, line 1: Not sure if you are follow the newline after header convention but this one is missing

reweeden: I was trying to add them in

mattp0 (dismissed):

mattp0 (approved):

PR #672, opened 2022-12-13
Update s3access.md

krobin10:
Proposing updating the token description fields cases to be in alignment with the updates requested by NASA.


reweeden:
Already updated on the `devel` branch. Thanks though

Issue #275, opened 2020-10-28
Added RequestId to cumulus format log messages


bbuechler:
seems obvious in retrospect, but would be useful for capturing the final result.

Alternative solution: 
1) detect a dict object being passed into a call `logger` call...
2) add each element of the dict to the log JSON... 
3) have `{"message": "<VALUE>"}`  be something to the affect of "cumulus log response payload"

https://github.com/asfadmin/thin-egress-app/blob/4111b52481a7694f88370b6dc64ed0a066fd9c0e/lambda/app.py#L140

bbuechler:
Released in https://github.com/asfadmin/thin-egress-app/releases/tag/tea-build.105

PR #677, opened 2023-01-02
Add some type hints and docstrings for app.py

yuvipanda:
I loved reading
https://github.com/asfadmin/thin-egress-app/blob/master/docs/vision.md! Hopefully this PR is useful and not just noise.

- I added type hints for returns and function parameters wherever I could determine them. 
   I couldn't figure out how exactly to type a return for the S3 botocore client, but otherwise
   I think this is accurate.
- I also added docstrings for the majority of functions after reading through the code to understand
   what it's doing.

### DIFF
--- a/lambda/app.py
+++ b/lambda/app.py
@@ -96,7 +96,7 @@ TEMPLATE_MANAGER = TemplateManager(conf_bucket, template_dir)
 
 
 @ttl_cache(maxsize=2, ttl=10 * 60, timer=time.time)
-def get_black_list():
+def get_black_list() -> dict:
     endpoint = os.getenv('BLACKLIST_ENDPOINT', '')
     if endpoint:
         response = urllib.request.urlopen(endpoint).read().decode('utf-8')
@@ -108,7 +108,7 @@ app = Chalice(app_name='egress-lambda')
 
 
 @app.middleware('http')
-def initialize(event, get_response):
+def initialize(event, get_response) -> Response:
     JWT_MANAGER.black_list = get_black_list()
     jwt_keys = retrieve_secret(os.getenv('JWT_KEY_SECRET_NAME'))
     JWT_MANAGER.public_key = base64.b64decode(jwt_keys.get('rsa_pub_key', '')).decode()
@@ -118,7 +118,7 @@ def initialize(event, get_response):
 
 
 @app.middleware('http')
-def set_log_context(event: chalice.app.Request, get_response):
+def set_log_context(event: chalice.app.Request, get_response) -> Response:
     origin_request_id = event.headers.get('x-origin-request-id')
 
     log_context(
@@ -139,7 +139,7 @@ def set_log_context(event: chalice.app.Request, get_response):
 
 
 @app.middleware('http')
-def forward_origin_request_id(event: chalice.app.Request, get_response):
+def forward_origin_request_id(event: chalice.app.Request, get_response) -> Response:
     response = get_response(event)
 
     origin_request_id = event.headers.get('x-origin-request-id')
@@ -268,7 +268,7 @@ def get_origin_request_id() -> Optional[str]:
 
 
 @with_trace()
-def get_aux_request_headers():
+def get_aux_request_headers() -> dict:
     req_headers = {"x-request-id": get_request_id()}
     origin_request_id = get_origin_request_id()
 
@@ -282,12 +282,12 @@ def get_aux_request_headers():
 
 
 @with_trace()
-def check_for_browser(hdrs):
+def check_for_browser(hdrs) -> bool:
     return 'user-agent' in hdrs and hdrs['user-agent'].lower().startswith('mozilla')
 
 
 @with_trace()
-def get_user_from_token(token):
+def get_user_from_token(token) -> Optional[str]:
     """
     This may be moved to rain-api-core.urs_util.py once things stabilize.
     Will query URS for user ID of requesting user based on token sent with request
@@ -405,7 +405,7 @@ def restore_bucket_vars():
 
 
 @with_trace()
-def do_auth_and_return(ctxt):
+def do_auth_and_return(ctxt) -> Response:
     log.debug('context: {}'.format(ctxt))
     here = ctxt['path']
     if os.getenv('DOMAIN_NAME'):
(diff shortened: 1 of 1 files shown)


yuvipanda:
Not sure if this should target master or devel?

reweeden:
Hey @yuvipanda, thanks for the PR! The way we have the repo set up right now, we make PR's into `devel` first, so if you could change the base branch that would be great.

yuvipanda:
Great! git macigck'd and base changed :)

mckadesorensen (approved):

yuvipanda:
@mckadesorensen sorry to change the PR after you had already reviewed it! I added some more docstrings as I read through the code to help me understand what was going on :) Hope that is useful too!

yuvipanda:
Also lmk what else I can do to help get this merged!

lambda/app.py
```diff
@@ -255,20 +293,32 @@ def get_success_response_headers(self) -> dict:
...
+    """
     assert app.lambda_context is not None
 
     return app.lambda_context.aws_request_id
 
 
 @with_trace()
 def get_origin_request_id() -> Optional[str]:
+    """
+    Return the *original* AWS Lambda Request ID
```
reweeden, line 307: I'm actually not sure what `x-origin-request-id` is. But I believe 'origin' has a specific meaning in HTTP and I think AWS is using it in a similar way.

lambda/app.py
```diff
@@ -367,12 +420,18 @@ def get_user_from_token(token):
 
 @with_trace()
 def cumulus_log_message(outcome: str, code: int, http_method: str, k_v: dict):
+    """
+    Emit log message to stdout in a format that cumulus understands
```
reweeden, line 424: Does anyone know if this actually interacts with cumulus at all?

reweeden (commented):
There's 1 line that becomes too long when adding the return type annotation. We'll need to fix that before being able to merge (see github actions).

Technically our coding standards say that docstrings should follow this format:
```
"""A multi line docstring has a one-line summary, less than 120 characters.

Then a new paragraph after a newline that explains in more detail any
general information about the function, class or method. Example usages
are also great to have here if it is a complex class or function.

When writing the docstring for a class, an extra line should be placed
after the closing quotations. For more in-depth explanations for these
decisions see <http://www.python.org/dev/peps/pep-0257/>

If you are going to describe parameters and return values, use Sphinx, the
appropriate syntax is as follows.

:param foo: the foo parameter
:param bar: the bar parameter
:returns: return_type -- description of the return value
:returns: description of the return value
:raises: AttributeError, KeyError
"""
```

This is straight from the OpenStack style guide.

lambda/app.py
```diff
@@ -24,6 +24,8 @@
     from opentelemetry import trace
     from opentelemetry.propagate import inject
 except ImportError:
+    # If opentelemetry is not installed, we make dummy objects / functions here
+    # so we do not need to add conditionls throughout our codebase
```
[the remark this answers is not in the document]
mckadesorensen: I agree with @reweeden on this comment. I have been burned too many times by people not updating comments when they update code.

lambda/app.py
mckadesorensen: @benbart Might know the answer to that.

lambda/app.py
```diff
@@ -450,7 +515,10 @@ def make_redirect(to_url, headers=None, status_code=301):
 
 
 @with_trace()
-def make_html_response(t_vars: dict, headers: dict, status_code: int = 200, template_file: str = 'root.html'):
+def make_html_response(t_vars: dict, headers: dict, status_code: int = 200, template_file: str = 'root.html') -> Response:
```
mckadesorensen, line 518: I think it would be best if we kept the line length to 120

lambda/app.py
benbart: More like cumulus interacts with the log message generated. This outputs the log message in a format that cumulus likes, if memory serves. The Issue #275 isn't very descriptive on this matter.

yuvipanda:
Thank you for the comments! I'm going to try to come back to this whenever I can, and do bits and pieces whenever I get the time :)

lambda/app.py
yuvipanda: @mckadesorensen fixed!

lambda/app.py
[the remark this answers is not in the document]
yuvipanda: My personal experience has been that the more common phenomena is lack of comments rather than wrong comments, and usually this 'mismatch' issue is dealt with during code review of the changing code itself rather than by not having comments at all.

And as for this specific pattern, I've personally seen a different pattern (detailed in https://adamj.eu/tech/2021/12/29/python-type-hints-optional-imports/) be used more often. It also has the advantage of working better with mypy and other type checkers, rather than simply setting to `None`. This was the primary reason I added the comment here.

However, y'all are the maintainers so are ultimately responsible for the state of the code. So I'll happily remove the comment if you think that's what's best for this project.

Thanks for your consideration!

lambda/app.py
yuvipanda: My understanding would be that there's some elasticsearch / kibana rule that is looking for log messages in this format to expose specially, and this is writing things out in that format. If you think a different kinda comment would be more clarifying here, I can do that.

yuvipanda:
@mckadesorensen re: docstrings, my hope is that adding *some* is better than none. I often find it difficult to write param dosctrings, as they often are repetitive. However, if you would like me to make sure all docstrings are 'complete', I can try (although would prefer not to haha).

reweeden:
Params in docstrings are optional. Just format what you have so it matches the example.

lambda/app.py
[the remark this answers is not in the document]
reweeden: Unfortunately I don't think code reviews can be relied upon to catch comment mismatches. Stuff like that slips through unnoticed all the time which is the nature of the problem.

I'd say, lets go for no comment in this case.

reweeden:
Hi @yuvipanda, are you still interested on working on this?

reweeden:
As I haven't seen activity from the author in over 1 year, I am going to close this PR. Feel free to re-open it if you wish to come back to this.

yuvipanda:
Thanks @reweeden! I haven't been able to dedicate any time to this in a while, and I doubt I'll be able to anytime soon. Closing feels right.

Hope to be able to collaborate with y'all in some other capacity somewhere :)

PR #705, opened 2023-03-01
Refactor GitHub Actions

reweeden:
Github has added some features to actions since these workflows were created that make a lot of things quite a bit simpler or nicer. In this PR I'm updating the workflows to make use of these new features.

1. Repository configuration variables. *Previously only secrets were available which caused a number of headaches since they are redacted in logs and the `secrets` context is restricted in where it can be used in the workflows.*
2. Inherit secrets when calling a reusable workflow. *Previously all secrets had to be listed explicitly which created some annoyingly long copy/pasta secret lists.*

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -11,11 +11,10 @@ jobs:
         with:
           python-version: 3.8
 
-      - run: pip install -r requirements/requirements.txt
-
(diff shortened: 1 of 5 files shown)


mattp0 (approved):

mattp0 (dismissed):
my bad

PR #708, opened 2023-03-03
Use moto for tests

reweeden:
I also refactored the static test data fixture to be the same as in our other projects.


tests/test_app.py
```diff
@@ -826,7 +844,9 @@ def test_logout(
 
 
 @mock.patch(f"{MODULE}.do_login", autospec=True)
```
benbart, line 846: Wasn't this `f"{MODULE}.do_login"` form a workaround for the situation where the directory was named `lambda`?

benbart (approved):
I'm going to steal some of this.

tests/test_app.py
reweeden: No, `mock.patch` always requires a string argument. The workaround was the usage of `importlib` to import the module rather than using an `import` statement. The `MODULE` variable was just a refactor of the import path which I was able to use both in the `mock.patch` calls and in the `importlib` call since both of those functions require the same string.

mattp0 (approved):

mckadesorensen (approved):

PR #729, opened 2023-06-05
PR-4559 Update S3 docs to include multipart_chunksize guidance

mattp0:
Update docs to include multipart_chunksize guidance in readme and update python example to use transfer_config object

### DIFF
--- a/docs/s3access.md
+++ b/docs/s3access.md
@@ -63,6 +63,10 @@ as the source bucket you are pulling from. See
 [Using temporary credentials with AWS resources](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html)
 for more information on how to use your temporary credentials.
 
+### Multipart Chunksize
+
+The AWS `multipart_chunksize` is 8MB by default. It is recommended to increase this to between 128MB-1GB as shown in the example below. Additional resources can be found in the [AWS documentation](https://docs.aws.amazon.com/cli/latest/topic/s3-config.html#configuration-values).
+
 **Example:**
 
 This example lambda function uses
@@ -77,6 +81,7 @@ import boto3
 import json
 import urllib.request
 
+MB = 1000 ** 2
 
 def lambda_handler(event, context):
     # Get temporary download credentials
@@ -94,10 +99,15 @@ def lambda_handler(event, context):
         "s3",
         aws_access_key_id=creds["accessKeyId"],
         aws_secret_access_key=creds["secretAccessKey"],
-        aws_session_token=creds["sessionToken"]
+        aws_session_token=creds["sessionToken"],
     )
     # Lambda needs to have permission to upload to destination bucket
     upload_client = boto3.client("s3")
+
+    # Set up transfer config object
+    transfer_config = {
+        multipart_chunksize = 128 * MB
+    }
 
     # Stream from the source bucket to the destination bucket
     resp = download_client.get_object(
@@ -105,9 +115,10 @@ def lambda_handler(event, context):
         Key=event["Source"]["Key"],
     )
     upload_client.upload_fileobj(
-        resp["Body"],
-        event["Dest"]["Bucket"],
-        event["Dest"].get("Key") or event["Source"]["Key"],
+        Fileobj=resp["Body"],
+        Bucket=event["Dest"]["Bucket"],
+        Key=event["Dest"].get("Key") or event["Source"]["Key"],
+        Config=transfer_config,
     )
 ```
 


docs/s3access.md
```diff
@@ -77,6 +81,7 @@ import boto3
 import json
 import urllib.request
 
+MB = 1024 ** 2
```
reweeden, line 84: Pretty sure
```suggestion
MiB = 1024 ** 2
```

docs/s3access.md
```diff
@@ -63,6 +63,10 @@ as the source bucket you are pulling from. See
 [Using temporary credentials with AWS resources](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html)
 for more information on how to use your temporary credentials.
 
+### Multipart Chunksize
+
+The AWS multipart_chunksize by default is 8MB, it is recommended that a multipart_chucksize of 128MB-1GB is used. This is included in the example below. Addtional resources can be found from [AWS](https://docs.aws.amazon.com/cli/latest/topic/s3-config.html#configuration-values)
```
reweeden, line 68: Is it MB or MiB?

reweeden (changes requested):

docs/s3access.md
mattp0: there is not direct guidance from AWS on what the unit is. I think in this case I did use a power of 2 for the 1024 so it should be MiB.

docs/s3access.md
mattp0: the docs are showing examples with MB. so I should swap my top to be 1000?. 1MB = 1000 Kb and 1MiB = 1024 Kib.

docs/s3access.md
mattp0: I updated it to be 1000 and use MB as the docs from AWS do.

docs/s3access.md
```diff
@@ -99,15 +104,21 @@ def lambda_handler(event, context):
     # Lambda needs to have permission to upload to destination bucket
     upload_client = boto3.client("s3")
 
+    # Set up transfer config object
+    transfer_config = {
+        multipart_chunksize= 128 * MB
```
reweeden, line 109: PEP 8
```suggestion
        multipart_chunksize = 128 * MB
```

docs/s3access.md
```diff
@@ -63,6 +63,10 @@ as the source bucket you are pulling from. See
 [Using temporary credentials with AWS resources](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html)
 for more information on how to use your temporary credentials.
 
+### Multipart Chunksize
+
+The AWS `multipart_chunksize` by default is 8MB, it is recommended that a `multipart_chuncksize` of 128MB-1GB is used. This is included in the example below. Additional resources can be found from [AWS](https://docs.aws.amazon.com/cli/latest/topic/s3-config.html#configuration-values)
```
reweeden, line 68: Still misspelled
```suggestion
The AWS `multipart_chunksize` by default is 8MB, it is recommended that a `multipart_chunksize` of 128MB-1GB is used. This is included in the example below. Additional resources can be found from [AWS](https://docs.aws.amazon.com/cli/latest/topic/s3-config.html#configuration-values)
```

reweeden, line 68:
```suggestion
The AWS `multipart_chunksize` by default is 8MB, it is recommended that a `multipart_chuncksize` of 128MB-1GB is used. This is included in the example below. Additional resources can be found from [AWS](https://docs.aws.amazon.com/cli/latest/topic/s3-config.html#configuration-values).
```

reweeden, line 68: I find some of the phrasing/punctuation to be a bit awkward. Maybe something like this would be better. I also wonder if we should be explaining why the bigger limit is better?
```suggestion
The AWS `multipart_chunksize` is 8MB by default. It is recommended to increase this to between 128MB-1GB as shown in the example below. Additional resources can be found in the [AWS documentation](https://docs.aws.amazon.com/cli/latest/topic/s3-config.html#configuration-values)
```

reweeden (changes requested):
Another thing to consider is if this change actually belongs in TEA or if it would be better made to our sentinel1 templates. Do we actually know that the larger chunk size is better for ALL DAACs? Or is this just an optimization for our Sentinel archive because we distribute everything as one large zip file? We have to make sure that the changes that make it into this repository would be correct for any DAAC that might enable s3 direct access through TEA, now and in the future.

mattp0:
> Another thing to consider is if this change actually belongs in TEA or if it would be better made to our sentinel1 templates. Do we actually know that the larger chunk size is better for ALL DAACs? Or is this just an optimization for our Sentinel archive because we distribute everything as one large zip file? We have to make sure that the changes that make it into this repository would be correct for any DAAC that might enable s3 direct access through TEA, now and in the future.

In my understanding this is a challenge that all DAACs will face as this request relates directly to lowering the number of logs generated for distribution metrics. As the file size was too large for the team to ingest on the NASA side of the house. I think it makes sense to put in TEA and maybe I should add some extra text about the rationale?

PR #734, opened 2023-07-07
PR-4061 Add TEA command line tool

reweeden:
The idea is that this will replace the deploy.sh script but also include other standalone functionality.

The best way to review would be to install it and try it out.


tea-cli/tea_cli/deploy.py
```diff
@@ -0,0 +1,282 @@
...
+            )
+
+
+class CloudFormationStack(Resource):
+    def __init__(
+        self,
+        stack_name: str,
+        template_body: str = None,
+        template_url: str = None,
+        parameters: dict = {},
```
benbart, line 112: Is it OK to set a mutable default now?

reweeden: IMO it's fine so long as you don't mutate it. My philosophy on this is that the real problem that should be avoided is performing mutating operations on input arguments. Arguments should be considered 'read only'. If you need some sort of mutable persistent state then you should be writing a class and using member variables for that.

setup.cfg
```diff
@@ -1,6 +1,5 @@
 [isort]
-multi_line_output = 3
-line_length = 120
+profile = black
```
mckadesorensen, line 2: A Google search shows that blacks defalut is 88 characters per line. Our coding standard states 120.

reweeden: Yep and 88 < 120 so using the black profile for isort will still ensure that our import formatting always conforms to our coding standards.

tea-cli/tea_cli/commands/quick_deploy.py
```diff
@@ -0,0 +1,473 @@
...
+def configure_parser(parser: argparse.ArgumentParser):
+    parser.add_argument("--edl-uid", help="EDL app uid", dest="edl_uid")
+    parser.add_argument("--edl-pass", help="EDL app password", dest="edl_password")
+    parser.add_argument("--edl-client-id", help="EDL app client ide", dest="edl_client_id")
+
+
+def handle_args(args: argparse.Namespace):
+    quick_deploy(
+        stack_name="reweeden-tea-quickdeploy-test",
+        profile_name="rain-dev",
```
reweeden, line 53: Need to pull these from command line parameters

setup.cfg
mckadesorensen: 🤣  I think the `line_length = 120` needs to stay

tea-cli/tea_cli/commands/list_versions.py
```diff
@@ -0,0 +1,130 @@
...
+    return parser
+
+
+def configure_parser(parser: argparse.ArgumentParser):
+    pass
+
+
+def handle_args(args: argparse.Namespace):
+    list_versions(
+        profile_name="rain-dev"
```
reweeden, line 104: Same here

setup.cfg
reweeden: Not a joke. There is no line minimum, there is simply a max of 120. The black profile still conforms to that. Looking at our other repos there are a number of them that are configured to break at 80. Keep in mind this is only for automated formatting of imports, this setting does not affect the flake8 line length for the whole file which is set at 120.

benbart:
I wonder if we'll run into collisions with other command line tools called tea. Like: https://docs.tea.xyz/getting-started/using-tea

tea-cli/tea_cli/commands/quick_deploy.py
benbart: yes

tea-cli/tea_cli/deploy.py
```diff
@@ -0,0 +1,282 @@
...
+
+
+class Secret(Resource):
+    def __init__(self, name: str, description: str, secret_string: str):
+        self.name = name
+        self.description = description
+        self.secret_string = secret_string
+
+    def deploy(self, session: boto3.Session):
+        client = session.client("secretsmanager")
```
benbart, line 186: I get a `botocore.exceptions.NoRegionError: You must specify a region.` here. IMO, we should default to `us-west-2`

reweeden:
Hmmm. I suppose `teacli` wouldn't be much worse to type. I don't really want to be typing `thin-egress-app` though...

benbart:
Or a more collision-resistant `awstea`?

tea-cli/tea_cli/deploy.py
reweeden: Hmm, I have the region set in my aws profile, I think because otherwise you get this error even when running just regular aws cli commands

reweeden: I guess, do you get that error if you try to use the existing deploy.sh script?

benbart: Strangely, it didn't complain here: https://github.com/asfadmin/thin-egress-app/pull/734/files#diff-a6fc357f8c1126d7a4a3252de5826111152f47bd209789bfe38545c48ae33eceR46

tea-cli/tests/test_main.py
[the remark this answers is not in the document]
reweeden: I don't think so really. So far there isn't really any code that doesn't interact with the console IO or perform some boto3 call that would be easy to test. I was thinking there might be some generic machinery code that could be tested, but there really isn't any.

mattp0 (approved):

PR #740, opened 2023-08-10
JWT script improvements

benbart:


gjclark (approved):

scripts/setup_jwt_cookie.sh
```diff
@@ -14,7 +14,7 @@ EOL
 
 function GENERATE_TEA_CREDS {
   cd /tmp || exit 1
-  ssh-keygen -t rsa -b 4096 -m PEM -f ./jwtcookie.key -N ''
+  ssh-keygen -t rsa -b 4096 -m PEM -f ./jwtcookie.key -N '' -C "asfjwt"
```
reweeden, line 17: How about writing a complete sentence like 'A key used by Thin Egress App to sign JWTs'. I'm not sure we want to put `ASF` in there since this may be used by other DAACs.

benbart: I agree about putting the ASF in there.

PR #769, opened 2023-11-09
PR-5371 Update all dependencies

reweeden:
The mkdocs ASF theme was updated in https://github.com/ASFHyP3/mkdocs-asf-theme/issues/27 to support the latest version of mkdocs material, so this should address the warnings about the outdated version of mkdocs material.

### DIFF
--- a/requirements/requirements-dev.txt
+++ b/requirements/requirements-dev.txt
@@ -4,33 +4,35 @@
 #
 #    pip-compile requirements/requirements-dev.in
 #
-boto3==1.26.153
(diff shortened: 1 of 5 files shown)


mattp0 (approved):
Seems like it was straight forward

mattp0:
WRT WhiteSource can we got to PIP 23.3?

reweeden:
I don't know where it even thinks its getting a version of pip from. Pip isn't pinned so it's probably just whatever comes with the python3.8 dockerfile.

PR #777, opened 2023-12-19
PR-5069 Refactor locate endpoint

reweeden:
The `/locate` endpoint seems to be the last place that had special handling code to interact with the bucket map that was not going through the `BucketMap` class. This change will ensure that the bucket map behavior is consistent for `/locate` and other functionality. Specifically, this allows the old-style bucket maps with no top level `MAP` key to work with `/locate` (previously a 500 error would be returned).

### DIFF
--- a/requirements/requirements.in
+++ b/requirements/requirements.in
@@ -1,6 +1,5 @@
 cachetools
 cfnresponse
 chalice
-flatdict
 git+https://github.com/asfadmin/rain-api-core.git@318aac226c92cf6f60cc8821d6d94669485972c6
 netaddr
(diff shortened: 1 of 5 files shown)


mattp0 (approved):

mckadesorensen (approved):
Assuming it works, it looks good to me.

mckadesorensen:
@reweeden It is failing an automatic check.

gjclark (approved):

reweeden:
> @reweeden It is failing an automatic check.

It's pyup complaining about a dependency. I'm going to ignore it for 2 reasons:
1. This PR is not about updating dependencies
2. The dependency is required to be at that version by chalice

PR #803, opened 2024-03-19
PR-5790 Add CORS preflight OPTIONS request

reweeden:
Some libraries will automatically make this `OPTIONS` request and we need to respond with the appropriate headers to prevent cross origin requests from being blocked.


mattp0 (approved):
Nice! That was pretty quick and not too many changes. 

Should we be concerned about the decrease in code cov here? Seems like there could be some easy unit tests added.

mckadesorensen (approved):

reweeden:
Yea I just added a some unit tests.

PR #821, opened 2024-06-11
Use aws_s3_object instead of aws_s3_bucket_object

### LINKED ISSUE #801: update Terraform code to prevent Deprecated Resource warnings
When deploying TEA, I get the following Terraform deprecation warning:

```
Warning: Deprecated Resource

  with module.thin_egress_app.aws_s3_bucket_object.lambda_source,
  on .terraform/modules/thin_egress_app/main.tf line 25, in resource "aws_s3_bucket_object" "lambda_source":
  25: resource "aws_s3_bucket_object" "lambda_source" {

use the aws_s3_object resource instead
```

gjclark:
Closes #801

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,6 @@
 
 # Tools
 .terraform
-/terraform/providers.tf
 terraform.tfstate*
 *.tfvars
 /thin_egress_app/.chalice
--- a/terraform/providers.tf
+++ b/terraform/providers.tf
@@ -1,0 +1,8 @@
+terraform {
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = ">= 4.0"
+    }
+  }
+}
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -22,21 +22,21 @@ resource "aws_s3_bucket" "lambda_source" {
   tags = merge(var.tags, { DAR = "YES" })
 }
 
-resource "aws_s3_bucket_object" "lambda_source" {
+resource "aws_s3_object" "lambda_source" {
   bucket = aws_s3_bucket.lambda_source.bucket
   key    = "${filemd5(local.lambda_source_filename)}.zip"
   source = local.lambda_source_filename
   etag   = filemd5(local.lambda_source_filename)
 }
 
-resource "aws_s3_bucket_object" "lambda_code_dependency_archive" {
+resource "aws_s3_object" "lambda_code_dependency_archive" {
   bucket = aws_s3_bucket.lambda_source.bucket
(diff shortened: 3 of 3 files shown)


mckadesorensen (approved):

mattp0 (approved):

reweeden (commented):
I wonder if it would be a good idea to add a `required_providers` block to the terraform module so we can pin the minimum supported version of the AWS provider. I'm actually a little surprised that it works without that.

gjclark:
> I wonder if it would be a good idea to add a `required_providers` block to the terraform module so we can pin the minimum supported version of the AWS provider. I'm actually a little surprised that it works without that.

NISAR has:
```
terraform {
  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = "~> 5.0"
    }
    null = {
      source  = "hashicorp/null"
      version = "~> 2.1"
    }
  }
}
```
Should I just duplicate that here?

reweeden:
You should only include the providers that the module actually needs and you should use `>=` as explained on the best practices section here: https://developer.hashicorp.com/terraform/language/providers/requirements#best-practices-for-provider-versions

.gitignore
```diff
@@ -11,7 +11,6 @@
 
 # Tools
 .terraform
-/terraform/providers.tf
```
reweeden, line 14: Lol, why was that in there?

terraform/providers.tf
```diff
@@ -0,0 +1,8 @@
+terraform {
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = ">= 5.0"
+    }
+  }
+}
```
reweeden, line 8: Missing newline

terraform/providers.tf
```diff
@@ -0,0 +1,8 @@
+terraform {
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = ">= 5.0"
```
reweeden, line 5: This isn't the minimum. I just checked a random 4.x version and 4.63 has the `aws_s3_object` already. So you can go down at least to that.

You should find the lowest version that has that resource. I would do it through trial and error using this page and selecting the version from the version dropdown:
https://registry.terraform.io/providers/hashicorp/aws/4.63.0/docs/resources/s3_bucket_object

.gitignore
gjclark: Hah I don't know. I was hoping to find some discussion in the PR that brought that in. Couldn't find the reason.

terraform/providers.tf
gjclark: Looks like 4.0 introduced it.

reweeden (approved):

PR #839, opened 2024-09-09
Update Dependencies

reweeden:


gjclark (approved):

mckadesorensen (approved):

mckadesorensen:
Looks like these updates are failing some checks.

reweeden:
> Looks like these updates are failing some checks. Other than that, the code updates look good.

It's just vulnerability scanner noise. Pyup complains that the latest version of Jinja2 has a "vulnerability", at the same time:
> NOTE: The maintainer and multiple third parties believe that this vulnerability isn't valid because users shouldn't use untrusted templates without sandboxing.

PR #844, opened 2024-10-03
Fix Terraform updating dependency layer object when content hasn't changed

reweeden:
While doing Cumulus upgrades, I noticed that on every update Terraform was adding something like this to the plan:

```
 # module.thin_egress_app.aws_s3_bucket_object.lambda_code_dependency_archive will be updated in-place
  ~ resource "aws_s3_bucket_object" "lambda_code_dependency_archive" {
      ~ etag                          = "661d0e0eb0bc955857a7a71087a4208a-2" -> "96bac3b47b39d00107729959bc6689d3"
        id                            = "96bac3b47b39d00107729959bc6689d3.zip"
        tags                          = {}
      + version_id                    = (known after apply)
        # (20 unchanged attributes hidden)
    }
```

The [etag](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_object#etag) attribute has a warning about using it for files larger than 16MB as it can cause this behavior where the etag after uploading with a multipart upload does not match the full source code hash causing Terraform to needlessly update the object.

While I was at it I also applied the `tags` variable to the bucket objects as well.

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -27,6 +27,7 @@ resource "aws_s3_object" "lambda_source" {
   key         = "${filemd5(local.lambda_source_filename)}.zip"
   source      = local.lambda_source_filename
   source_hash = filemd5(local.lambda_source_filename)
+  tags        = var.tags
 }
 
 resource "aws_s3_object" "lambda_code_dependency_archive" {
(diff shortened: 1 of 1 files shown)


mattp0:
"Triggers updates when the value changes."

Won't this cause updates not to occur when the content changes ?

mattp0 (approved):

gjclark (approved):

PR #858, opened 2025-01-06
Add error handling for malformed auth headers

reweeden:
Currently if we send an auth header without the string `Bearer` we'll see an error like this:
```
$ curl -X GET --header "Authorization: foobar" http://...
{"Code":"InternalServerError","Message":"An internal server error occurred."}
```

With these changes:
```
$ curl -X GET --header "Authorization: foobar" http://...
{"status_code":400,"error_description":"Malformed Authorization header"}
```

NOTE: CMR accepts an authorization header containing just the token, so it is reasonable that some users might see that and try the same approach with TEA.

### DIFF
--- a/thin_egress_app/app.py
+++ b/thin_egress_app/app.py
@@ -180,10 +180,14 @@ class RequestAuthorizer:
             self._response = do_auth_and_return(app.current_request.context)
             return None
 
-        method, token, *_ = authorization.split()
-        method = method.lower()
+        parsed_auth = self._parse_auth_header(authorization)
+        if parsed_auth is None:
+            return None
+
+        method, args = parsed_auth
 
         if method == "bearer":
+            token, *_ = args
             # we will deal with "bearer" auth here. "Basic" auth will be handled by do_auth_and_return()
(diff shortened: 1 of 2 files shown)


gjclark (approved):

mckadesorensen (approved):

thin_egress_app/app.py
```diff
@@ -203,6 +207,32 @@ def get_profile(self) -> Optional[UserProfile]:
...
+        except Exception:
+            if check_for_browser(app.current_request.headers):
+                template_vars = {
+                    "title": "Bad Request",
+                    "status_code": 400,
+                    "contentstring": "Malformed Authorization header",
+                    "requestid": get_request_id(),
+                }
+
+                self._response = make_html_response(template_vars, {}, 403, "error.html")
```
mattp0, line 224: intentionally checking for a 400 then making a 403 response?

mattp0 (approved):

thin_egress_app/app.py
reweeden: Oops

PR #874, opened 2026-02-18
PR-7323 update dependencies for TEA

mattp0:
## Pull Request Checklist

I have:
- [x] performed a self review of my code [I&A code style](https://github.com/asfadmin/ia-standards/blob/main/standards.md)
    - Resources and Data Structures are sorted by ABC or a defined sorting pattern
- [x] updated the documentation accordingly
- [x] verified required action checks are passing
- [x] deployed my code and all new and existing E2E/Integration tests passed
- [x] bumped the version number as appropriate


Makefile
```diff
@@ -283,7 +283,7 @@ requirements/requirements-dev.txt: \
 	requirements/requirements.txt
 
 requirements/%.txt: requirements/%.in
-	$(DOCKER_DEPENDENCY_BUILDER) pip-compile -q -U --cache-dir /var/task/$(DIR)/.pip-cache/ $<
+	$(DOCKER_DEPENDENCY_BUILDER) pip-compile -q -U --cache-dir /var/task/$(DIR)/.pip-cache/ --no-strip-extras --output-file $@ $<
```
mattp0, line 286: Added the output file and added --no-strip-extras because of this warning:
>WARNING: --strip-extras is becoming the default in version 8.0.0. To silence this warning, either use --strip-extras to opt into the new default or use --no-strip-extras to retain the existing behavior.
docker run --rm --user "504:20" -v "$PWD":/var/task  --entrypoint "" tea-dependency-builder pip-compile -q -U --cache-dir /var/task/dist/.pip-cache/ --output-file requirements/requirements-test.txt requirements/requirements-test.in

reweeden: Hmm, interesting. I think stripping extras is probably fine. I don't think it would make a difference to the resulting .txt file since the whole dependency tree is already expanded out in there anyway. Looks like `uv` maybe also strips them by default.

src/tea-cli/pyproject.toml
reweeden: Bump version?

reweeden (approved):

src/tea-cli/pyproject.toml
reweeden: Nvm, you did it right after I made the comment haha

Makefile
mattp0: I only picked no-strip as it said thats the existing behavior. I could remove it and let it switch to the new default?

